PR #903, opened 2015-03-06
Improved detection of co-pilot seat for helicopters

nsgundy:
Add function TFAR_fnc_getCopilot to improve detection of units in co-pilot seat.
The previous implementation was only checking for units in turret [0],
but a lot of community created helicopter models do not have the co-pilot at that position,
resulting in the helicopter radio not being available to the co-pilot, but instead to one of the door gunners (just an example).
Examples are: HAFM (UH1Y Venom, CH-47F), RHS (UH-60M), Kyro (MH-47E)


nsgundy:
Just in case you are wondering about the performance impact: I measured an increased execution time of `TFAR_fnc_VehicleLR` from `0.113955 ms` to `0.160825 ms`.

michail-nikolaev:
Thanks, will check

nsgundy:
Just to give a reason for why I am restricting the improved detection to helicopters: I did check with the popular F/A-18F and for some reason they do not have the WSO (weapon systems officer / co-pilot) seat defined as a co-pilot set (you cannot take controls). I am currently trying to figure out if this is intentional [here](http://forums.bistudio.com/showthread.php?164308-F-A-18-Super-Hornet-and-Su-35S-Flanker-E&p=2892908&viewfull=1#post2892908).
Haven't checked for mods like the AC-130, but since the check for anything but helicopters hasn't changed, it is just as good / bad as before.

Checked with all vanilla, HAFM, RHS and Kyro helicopters and it seems to work properly. What I don't know is if there are any other checks for `turret [0]` that aren't using `turretUnit [0]` in TFAR. If so, I will have missed them.

dedmen:
Please merge 1.0 into your PR.

nsgundy:
Let me know if there is anything for me to do... been out of the loop for quite a while now.

kavinsky:
@nsgundy can you merge 1.0 branch in your PR?

nsgundy:
Should be mergable now. _Note_: I did not retest this (don't have things set up at the moment) and it is possible that since my implementation there are other places where you look up the co-pilots turret position that would benefit from using the new function. Please also have another shot at the new function to see if since its implementation there have been new SQF commands that it could potentially benefit from (mostly to further increase performance).

commy2:
A suggestion.
If the function is used often, then it should cache the vehicles turret path. This way it does not have to repeat the same config operations.

```sqf
private _turretPath = _this getVariable "TFAR_turretPath";

if (isNil "_turretPath") then {
    // do magic
    // _turretPath = ...
    _this setVariable ["TFAR_turretPath", _turretPath];
};

_this turretUnit _turretPath;
```

commy2:
I think the fastest way to report a vehicles co pilot turret I found was:
```
fullCrew [_vehicle, "turret", true] apply {_x select 3} select {
    getNumber ([_vehicle, _x] call CBA_fnc_getTurret >> "isCopilot") == 1
} param [0, []] // return
```
https://github.com/acemod/ACE3/blob/master/addons/common/functions/fnc_getTurretCopilot.sqf

dedmen:
@kavinsky will need to change the function for his #990 anyway so we can still do optimizations when he does that. So I'd approve this PR. Leave it to @kavinsky to merge and remember to do the optimizations later.

Issue #919, opened 2015-04-07
Update keybinding calls for new CBA


CorruptedHeart:
CBA's keybinding functions are going to change when Marksman DLC releases.

I will handle the change of our calls to it.

PR #920, opened 2015-04-07
Switched over to the new CBA's keybinding method.

CorruptedHeart:
Solves https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/919
Requires the latest Dev version of CBA.


michail-nikolaev:
Hm... Strange, I was thinking I have fixed it already

Issue #868, opened 2014-11-17
Event for radio dialog opening


michail-nikolaev:

CorruptedHeart:
This is to simply fire an event when a radio dialog has opened correct?

What sort of parameters would be necessary?

I'm thinking

- Radio (string/array)
- LR (bool if it is a LR radio)
- The dialog (a reference to the dialog, if the user wants to do something with it)
- Opened (bool if the dialog has opened or closed)

michail-nikolaev:
@CorruptedHeart agreed

CorruptedHeart:
Event: OnRadioOpen
Parameters Passed:
0: OBJECT - unit (player)
1: STRING/ARRAY - radio
2: BOOL - LR radio
3: STRING - dialog name
4: BOOL - dialog open

PR #922, opened 2015-04-07
Added OnRadioOpen events to all radios.

CorruptedHeart:
It fires when the radio dialog is opened and when it is closed.

Event: OnRadioOpen
Parameters Passed:
0: OBJECT - unit (player)
1: STRING/ARRAY - radio
2: BOOL - LR radio
3: STRING - dialog name
4: BOOL - dialog open

https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/868


michail-nikolaev:
Cool, thanks!

PR #925, opened 2015-04-09
ILD implementation

BaerMitUmlaut:
Interaural level differences (one channel is louder than the other) are one of the two parts that make audio directional and clunk does not seem to apply it to the audio stream.

This is a very basic ILD implemenation which I used myself for a while now and it helps quite a bit with localizing persons that are talking to you.

### DIFF
--- a/ts/src/plugin.cpp
+++ b/ts/src/plugin.cpp
@@ -762,6 +762,7 @@ void playWavFile(uint64 serverConnectionHandlerID, const char* fileNameWithoutEx
 				}
 				clientData->getClunk(id)->process(input, wave->_spec.channels, samples, position, clientData->viewAngle);
 				clientData->removeClunk(id);
+				applyILD(input, wave->_spec.channels, samples, position, clientData->viewAngle);
 			}
 
 		}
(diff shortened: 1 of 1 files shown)


michail-nikolaev:
Hm.. Interesting. Clunk should do it as well, but need to check

BaerMitUmlaut:
I just found out about this because my group was testing Acre 2 and TFAR and I was wondering why the direct chat was still directional, but somehow not as precise and this seems to be the reason.

michail-nikolaev:
is it hard task to create video which compares new and old code? (unable to test myself currently)

BaerMitUmlaut:
I don't have any video for comparison, but I made a quick sound recording. I tested by playing a sinus curve via a TS3 soundboard and playing it with a dropped radio with speakers turned on.

The upper recording is with my DLL, the bottom one the standard 0.9.7 DLL.

tfar dir test

Sound recording (normal first, then with ILD): http://vocaroo.com/i/s0moZns6AZyC

BaerMitUmlaut:
Forgot to say, whilst recording I just stood still and made a full rotation.

PR #947, opened 2015-05-03
Changeable Plugin filenames

### LINKED ISSUE #945: Plugin does not work if changing dll filename
If I change the filename of the Taskforce dll from task_force_radio_win64.dll to task_force_radio_3_0_16_win64.dll it doesnt receive plugin commands anymore.
Thats because in ts3plugin_onPluginCommandEvent in plugin.cpp 
the pluginName equals the filename 
and in Line 3044 its checked to be the exact name... 
It would be better to check if it starts or contains task_force_radio.
A fix would be to replace:
```c++
    if ((strcmp(pluginName, PLUGIN_NAME) == 0) || (strcmp(pluginName, PLUGIN_NAME_x64) == 0) || (strcmp(pluginName, PLUGIN_NAME_x32) == 0))
```
with
```c++
    if (strncmp(pluginName, PLUGIN_NAME, strlen(PLUGIN_NAME)) == 0)
```
For reference:
http://forum.teamspeak.com/showthread.php/61660-plugin-commands-are-tied-to-DLL-name

dedmen:
Fixes #945


michail-nikolaev:
Thanks!

PR #948, opened 2015-05-03
Customizable metaData

dedmen:
Allows user to customize the metaData status message without messing up "isPluginEnabledForUser"

### DIFF
--- a/ts/src/plugin.cpp
+++ b/ts/src/plugin.cpp
@@ -1256,7 +1256,8 @@ std::string getConnectionStatusInfo(bool pipeConnected, bool inGame, bool includ
 	std::string addon = serverIdToData[id].addon_version;
 	LeaveCriticalSection(&serverDataCriticalSection);
 
-	std::string result = std::string("[I]Connected[/I] [B]")
+	std::string result = std::string("TFAR")+ (pipeConnected ? "Y" : "N") + (inGame ? "Y" : "N")
+		+"[I]Connected[/I] [B]"
 		+ (pipeConnected ? "Y" : "N") + "[/B] [I]Play[/I] [B]"
 		+ (inGame ? "Y" : "N")
 		+ (includeVersion ? std::string("[/B] [I]P:[/I][B]") + PLUGIN_VERSION + "[/B]" : "")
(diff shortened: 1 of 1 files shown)


michail-nikolaev:
looks like it is not correct... what do you want to achive?

dedmen:
I made a fork where i changed the metaInfo in teamspeak.
Where i replaced the old getConnectionStatusInfo with
```c++
std::string getConnectionStatusInfo(bool pipeConnected, bool inGame, bool includeVersion)
{
	uint64 id = ts3Functions.getCurrentServerConnectionHandlerID();
	EnterCriticalSection(&serverDataCriticalSection);
	std::string addon = serverIdToData[id].addon_version;
	LeaveCriticalSection(&serverDataCriticalSection);
	std::string result = td::string("Connected: ")
 		+ (pipeConnected ? "Yes" : "No")
		+ "\nIn-Game: " + (inGame ? "Yes" : "No")
		+ (includeVersion ? "\nAddon-Version: " + ((addon.length() > 0) ? addon : "Unknown") : "")
	+(includeVersion ? (std::string("\nPlugin-Version: ") + PLUGIN_VERSION) : "");
	return result;
}
```
Which Works fine with people who use the same plugin... But people that play with the original plugin cant hear me because the original plugin checks for a big part of the Info text in isPluginEnabledForUser.
In this pull request its changed so that isPluginEnabledForUser only checks the first 6 chars which contain an identifier for TFAR and the data that isPluginEnabledForUser needs. That makes it possible to change the rest of the string in getConnectionStatusInfo without getting incompatible with other Plugin variants.
But that change would make older plugin versions incompatible :/ Maybe you should wait with merging till the next major release.

michail-nikolaev:
Hm... But why you fork your plugin and want to work with vanilla at the same time?

dedmen:
Im using my fork because i think it looks better but other people i play with dont use my version so its incompatible and i cant play with them.

dedmen:
This allows people to change their tfar info string in their own fork and still be able to play with vanilla tfar users. Though as probably no one besides me ever did that this is kinda useless. I'm not even sure if this still works so im just gonna close it.
Though i do prefer the longer version of the info thingy because i just think it looks nicer. Maybe that could be redone in a nicer style some day.

PR #954, opened 2015-05-21
Replaced logic of onSwTangentPressedHack function…

### LINKED ISSUE #845: LR activates SR
When speaking on (or trying to) on a LR radio, SR also gets activated. When you start speaking and during speaking you can see the hint saying LR, but when you terminate you see SR hint displaying for a brief second, which indicates it was activated (and not because of releasing CTRL before CAPS). You can also clearly hear yourself.

Version: 0.9.7.1
Keybinds: Default

This was not reproducible though, it only happened once and without server restart after all of us disconnected and 2-3 of us went back on (again without mission restart) and we didn't get any issues anymore.

GRUBES:
Rewrote onSwTangentPressedHack function with approach that strictly looks to determine that SWTransmit key is being pressed (including correct modifier keys) rather than trying to determine that LRTransmit key is not being pressed.

new approach should handle the cases where SWTransmit does or does not use any modifier keys, old code did not handle an unmodified LRTransmit correctly;
added method documentation

Fixes #845


michail-nikolaev:
One question - if I press CTRL+CAPS for LR and then release CTRL - will it start to transmit on SW?

GRUBES:
@michail-nikolaev I believe that yes, it will start to transmit on SW because `ClientInit.sqf` attaches `onSwTangentPressedHack` to the Key Up event. When you release CTRL, Hack function will be called and you will be pressing CAPS with no modifiers, so SW will transmit.

Is this the behaviour you want or don't want? This is sort of the behaviour I would expect.

I modify my TFAR keybindings, so when I was testing I tested with SWTransmit as CAPS and LRTransmit as TAB.

GRUBES:
Updated again; had declared `_keybind_sw` initially but then all references were `_sw_keybind`.

@michail-nikolaev Still curious about your last comment. If you press CTRL+CAPS to transmit on LR, then you release CTRL, what do you expect to happen?

I would think that as the user, since you are now pressing just CAPS, you would expect to be transmitting on your SW. The way the code is written right now, I believe this is what will happen (I will test to verify). Is this what you expect to happen?

If you *don't* want the SW to transmit upon releasing CTRL, I think you could remove the Key Up handler in ClientInit.sqf that calls this Hack function, then it would only react to Key Down events.

michail-nikolaev:
@GRUBES it is not good behaviour - it results in lot of false mesages in SW.

GRUBES:
@michail-nikolaev Ok, then what do you think about removing the Key Up handler from ClientInit.sqf? Since the hack function would now only detect whether you are exactly pressing the SWTransmit key, we should no longer need to worry about Key Up, should we?

Th3Cr0w:
Could you please upload your "fix" (.pbo) cause i need it soon ish and am ok with your solution.
Baking it myself promped some strange errors.

GRUBES:
@michail-nikolaev Been a while; any update on how you'd like to proceed with this (if at all)? I'm still not sure of any drawbacks to removing the Key Up handler; seems like it is no longer a necessary check, but maybe I'm missing the original reason for adding it in the first place.

michail-nikolaev:
I just want make sure if you have CTRL+CAPS pressed and releasing CTRL - no short range radio activated. Not sure how to achive it with latest changes - need to think here.

GRUBES:
I guess my question is why does the Hack (and subsequently the Shortwave Event Handler) get called at all on a Key *Up* event? Seems like we could only listen for Key Down, and then this would work fine. The only reason that this code even gets called when releasing CTRL (or any other key) is because of the Key Up handler in the init function. Why do we need to listen for Key Up?

GRUBES:
I'm not sure if I have been misreading ClientInit.sqf this entire time or what, but it seems that the SW Pressed event handler actually *does not* get called on any Key Up events, so I don't believe that with this solution, pressing CTRL+CAPS then releasing CTRL will transmit on SW. I am about to do some testing with default keybindings right now. I will update with results shortly.

GRUBES:
Actually, in doing testing with latest `master` and all of my customizations removed, it appears that this is already the behaviour. With SWTransmit on CAPS, and LRTransmit on CTRL+CAPS, if I hold CTRL+CAPS, I transmit on LR as expected. If I then release CTRL and continue holding CAPS, I begin transmitting on SW.

GRUBES:
I've updated the Pull Request with some new logic in the LR and DD `ReleasedHack` functions. Here is a summary of the results of these changes:

1. The original issue that this PR is intended to fix is the fact that when LR Transmit is bound to a keybind with no modifiers (`SHIFT`, `ALT`, `CTRL`), when the player presses LR Transmit, s/he would also transmit on SW at the same time. This is addressed by the updated logic in `onSwTangentPressedHack`.
1. During testing, I came across a secondary issue where after releasing LRTransmit, the very next attempt to transmit on SW would fail every time. I believe this was due to `onLRTangentReleasedHack` not correctly calling `onLRTangentReleased` when LRTransmit was an unmodified key. The updated logic in both LR and DD `on*TangentReleasedHack` functions handles this situation.

These two fixes make it so that LRTransmit does not transmit over both radios, and so players can switch between LR and SW transmission without any missed transmissions.

**Known Issue**: When using a LRTransmit keybind that is simply the same as SWTransmit with a modifier key (as in the default controls of `CTRL+CAPS` and `CAPS`, respectively), if the player releases the modifier key but continues to hold the other key, then the player will begin transmitting on the SW after 0.5s. The 0.5s delay comes from the `if` at the beginning of `onSwTangentPressed`. _This behaviour is already present in a clean 0.9.8 test mission_, and thus is not addressed by this pull request. There is already protection in place in the SWTransmit Pressed function for very quick false transmissions. It is assumed that if the user continues holding SWTransmit longer than the 0.5s, then they intend to transmit on SW.

michail-nikolaev:
@GRUBES thanks for you work. Will test as soon as get some free time

dedmen:
I agree that we don't want SW to transmit if we don't also release caps after releasing CTRL. 
Because the player knows that releasing CAPS will stop the transmission. He wouldn't expect that he can stop the current transmission and start one on another Radio before he did what he thought was required to stop a transmission. 
Getting into human psychology here :D 
It may have been the default behavior but i would really like to see this changed. And as you are changing the logic anyway with your PR i would like to also see that change in this PR.

kavinsky:
Ill hold this PR till i can solve the behavior. Most probably after 1.0-refactor

dedmen:
This was implemented along the same principle but still quite different in 3704c9a531fd928a52e185ac1a8af5aa570e2e50

PR #979, opened 2015-07-19
Auto Detect Teamspeak Version and adjust PLUGIN_API_VERSION

dedmen:
When Teamspeak asks for plugin apiversion detect which Teamspeak version is running and send API version 19 or 20 depending on version.
Removes the dependency of different plugins for different teamspeak version As the API changes dont affect the tfar plugin.
i dont know if 3.0.14 or 3.0.15 also uses version 20.
Please correct it if you know more.

I Made this fix because all our Arma Players have synced Folders and they have different Teamspeak versions.
I tried just naming the plugins task_force_radio_3_13_1_win32.dll and task_force_radio_3_16_win32.dll
but they couldnt hear eachother although my last pullrequest should have fixed that.


michail-nikolaev:
Cool.

Could you please add you to "Thanks" section in README file?

dedmen:
Done.
I hope everything is correct.

michail-nikolaev:
Not sure if it will cause any issues with older version of TS (because mismatch of API method declarations)

dedmen:
Yes if the API functions actually mismatch it wont work.. but none of the changed functions between 19 and 20 are actually used... thats why you can just change the version define to 19 or 20 and it works fine on both. When adapting in future versions that may have to be accounted for so old versions have to be removed from the auto-detect

PR #990, opened 2015-08-07
Use the new param command

kavinsky:
Added basic support for `params`, most function parameters now use that command

PR #992, opened 2015-08-08
 Add keybinds for different speech volumes #887

### LINKED ISSUE #887: Add keybinds for different speech volumes
It would be nice if we could map direct speech to keys with a certain volume, e.g. a yelling button such as right ctrl and still use my normal push to talk as normal.

ACyprus:
Adds a set of 'modifier' keys that allow a transient change in speaking volume whilst they are held down. 

I've introduced one for yelling and one for whispering.

There are no default keybinds.

Holding down a modifier key while transmitting will change the speaking volume and display a relevant hint.

Releasing the modifier will remove the hint and reset the speaking volume to the previous setting.

NOTE: the key binds in configuration are currently called "Yelling Modifier" and "Whispering Modifier" - only because I couldn't think of a better phrase. I am open to suggestions - perhaps something that would place them next to "Change Speaking Volume" in the list.

I've tested this on dedicated and works well. Could do with some 'third party' testing.

Closes #887


ACyprus:
PS: should I make this feature available to curator players?

michail-nikolaev:
Sorry - on vacation now + lost my dig - will review probably tommorow

ACyprus:
Thanks, No rush. Enjoy your vacation!

PS - when you're back, if you could let me know how I might initiate direct speech from the plugin then I can change these from just modifier keys to actual send speech keys. (Unless you meant you need to update the plugin first)

dedmen:
Can you please implement use of the "param" command into your onSpeakVolumeModifierPressed/Released functions?

like in #990

dedmen:
just for reference...... It doesn't work correctly when you switch between the modifier keys as fast as you can...

PR #995, opened 2015-08-14
use new allPlayers command instead of allUnits

commy2:
https://community.bistudio.com/wiki/allPlayers

This should help with server FPS on missions with a large number of AIs.
I added the alive check, because allUnits doesn't return dead units. Not sure if thats needed.


michail-nikolaev:
as I remember there are some issues with allPlayer some time ago...

kavinsky:
closing since its merged

PR #996, opened 2015-08-14
Fixed BOOLEAN is not a valid typeName

MikeMatrix:
The parameters were always ignored, because it was getting compared to "BOOLEAN", which is not a valid typeName.


michail-nikolaev:
Thanks!

PR #999, opened 2015-08-16
fixed #997

### LINKED ISSUE #997: using callExtension with empty string causes apphang
to reproduce:
"task_force_radio_pipe" callExtension ""

dll seems to be not happy with empty strings

X39:


X39:
its "not tested" but with such small changes i dont expect them to not work ... thus no actual testing needed just "maybe" the coding does not fits your style :laughing:

Issue #536, opened 2014-04-21
Support for BWMod


michail-nikolaev:
http://forums.bistudio.com/showthread.php?171183-BWMod

PR #1007, opened 2015-08-19
Update ..._items/config.cpp

Feivel69:
Line 97: Replaced "tf_mr3000_bwmod" with "tf_rt1523g_big_bwmod" since BWMod units are located at NATO side.

### DIFF
--- a/arma3/@task_force_radio/addons/task_force_radio_items/config.cpp
+++ b/arma3/@task_force_radio/addons/task_force_radio_items/config.cpp
@@ -94,7 +94,7 @@ class CfgFactionClasses
   // support for BW
   class BWA3_Faction
   {
-    backpack_tf_faction_radio = "tf_mr3000_bwmod";
+    backpack_tf_faction_radio = "tf_rt1523g_big_bwmod";
   };
 };
 


michail-nikolaev:
mr3000 was used because it is German radio.... What are you think?

Feivel69:
I didn't know it was used for BWMod faction on purpose ( https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/536#issuecomment-41473964 ), I thought it was accidentally. 

In real life, I think the Bundeswehr is using radios SEM70/80/90, so for Arma3 realism, I guess, it doesn't matter if mr3000 or rt1523, so I think it would be better to follow the scheme of TFR to have OPFOR using mr3000 and BLUEFOR using rt1523? 

But since the code for the vehicle radios was moved to BWMod itself it has to be changed in both mods (TFR/BWMOD) anyway or am I wrong? Is it possible to transmit from mr3000 to rt1523? 

Well, in the end it's a very low priority request, maybe one of the BWMod creators stumbles in to give their opinion. Kind regards.

kavinsky:
I think we should remove all our third party integration in our side and leave they integrate tfar features the way they want.

dedmen:
@Feivel69 won't merge. Instead we will remove BWA3_Faction completely in 1.0 so you can implement that in BWA3's side of code.

dedmen:
@commy2

commy2:
Roger.

PR #1009, opened 2015-08-23
Removed newline characters (stringtable.xml)

Feivel69:
Removed newline characters (~2000 characters), so source code formation doesn't effect ingame representation.

Picture:
https://img3.picload.org/image/iwiocrl/crlf.png

Edit: Only d089a24 is relevant. :)


michail-nikolaev:
Thanks!

PR #1050, opened 2016-03-08
Optimized RVExt main procedure.

novichiv:
Arma pipe extension:
Optimized and beautified RVExtension procedure.
TeamSpeak plugin:
Cleaned up some trash, fixed some bugs and undefined behaviour.


michail-nikolaev:
Thanks!

Issue #1078, opened 2016-06-26
allUnits is not guaranteed to include allPlayers


aeroson:
TFAR gets list of all units from allUnits.
Positions of those units are then send to TS plugin.
If unit position is **not** send to TS plugin, TS plugin thinks unit has **disabled** TFAR.

Problem:
Sometimes your computer doesn't include some players in allUnits array.
Result:
You can hear some players that are ingame as if they had disabled TFAR.

Seems if there is alot of units, Arma 3 goes into some kind of optimization mode and stops sending info about units that are far away.

So if there are 400 units on server and 30 players.
New player joins and spawns far away from me.
This player will not be included in allUnits and TFAR will think this player has TFAR disabled, because it will not get info about his position.
Thus this player will sound as if he had TFAR disabled.

Running this script seemed to fix it:
```
[{
	{ player reveal _x } forEach allPlayers;
},"BIS_fnc_spawn",true] call BIS_fnc_MP;
```

Seems to recur after player (re)connects.
Seems to recur after player respawns.

Can possibly be caused by starting the server with: -bandwidthAlg=2
Can possible be caused by ALiVE doing some behind the scenes optimizations.
Possibly related to Killzone Kid's comment at https://community.bistudio.com/wiki/reveal

Can be fixed by enabling serious mode.

**Nonetheless code should be changed to cope with the fact that allUnits is not guaranteed to include allPlayers.**

michail-nikolaev:
Thanks for investigations

PR #1079, opened 2016-06-28
Fix: allUnits is not guaranteed to include allPlayers

aeroson:
Fix for: https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/1078

Iam not sure if this is the best way to ensure that we have all players in the array.
It's untested.

### DIFF
--- a/arma3/@task_force_radio/addons/task_force_radio/functions/fn_processPlayerPositions.sqf
+++ b/arma3/@task_force_radio/addons/task_force_radio/functions/fn_processPlayerPositions.sqf
@@ -27,6 +27,10 @@ if !(isNull (findDisplay 46)) then {
 				tf_nearPlayers = call TFAR_fnc_getNearPlayers;
 			};
 
+			_other_units = allUnits - allPlayers; // non player units
+			_other_units = _other_units + allPlayers; // add player units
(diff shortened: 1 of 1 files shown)


arma3/@task_force_radio/addons/task_force_radio/functions/fn_processPlayerPositions.sqf
```diff
@@ -27,7 +27,9 @@ if !(isNull (findDisplay 46)) then {
 				tf_nearPlayers = call TFAR_fnc_getNearPlayers;
 			};
 
-			_other_units = allUnits - tf_nearPlayers;
+			_other_units = allUnits - allPlayers; // non player units
+			_other_units = _other_units + allPlayers; // add player units
```
michail-nikolaev: hm... what is sense to do - and after + ?

aeroson: to prevent duplicates,
allUnits + allPlayers could returns some players two times

aeroson:
However, wouldn't it be better to use only allPlayers ?

michail-nikolaev:
@aeroson I was used it before, but it sometimes not include all players

PR #1092, opened 2016-07-25
Load the plugin automatically, unless the user manually disabled it.

overfl0:
I think that a user that installs the plugin either by copying the dlls or by creating a .ts3_plugin and executing it would expect it to be enabled on next TeamSpeak run.
Otherwise he would not install those plugin files at all.

Also this is one less step for people wanting to install the plugin, which also means one less potential problem for them.

### DIFF
--- a/ts/src/plugin.cpp
+++ b/ts/src/plugin.cpp
@@ -2377,7 +2377,7 @@ void ts3plugin_freeMemory(void* data) {
  * This function is optional. If missing, no autoload is assumed.
  */
 int ts3plugin_requestAutoload() {
-	return 0;  /* 1 = request autoloaded, 0 = do not request autoload */
+	return 1;  /* 1 = request autoloaded, 0 = do not request autoload */
 }
 
 /************************** TeamSpeak callbacks ***************************/


michail-nikolaev:
Not sure it is good for new users...

michail-nikolaev:
O, I see. It WILL request autoload.

overfl0:
@dedmen I read your comment on Intercept github thread and I didn't know where I could  reach you, so I decided to contact you here.

If you guys are going to release a new version of TFAR, please ensure the new .dll files are re-created, so they will contain this PR (as opposed to 0.9.11 which was released after this PR has been merged but did not update the .dll files)

dedmen:
@overfl0 you could join our slack if you want to communicate ^^ Github should have PMs... This PR is already in 0.9.12 which will be released weekend i think. We have a new build system that ensures to rebuild every DLL on every release. besides the Arma extension cuz of battleye.

overfl0:
@dedmen That's great news. I'm interested in the slack channel but I cannot find the slack address anywhere.

kavinsky:
https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/1116

Issue #956, opened 2015-05-29
Vulnerable userconfig


BaerMitUmlaut:
For more info, read here:
http://forums.bistudio.com/showthread.php?192187-Userconfig-Vulnerabilities&p=2943475#post2943475

michail-nikolaev:
@BaerMitUmlaut thanks, will check, We did some tests here, but need to recheck.

kavinsky:
closing in favor of #1099

PR #1099, opened 2016-08-20
CBA Settings implementation to replace removed userconfig file

zgmrvn:
Here it is, it was pretty simple since the CBA Settings design fits perfectly the way you have designed the TFAR. The CBA Settings init function has to be executed on client and server (no matter your variable is server or client specific) so i have created a cba_settings.sqf file then included it in both of the init functions.


dedmen:
One question i have is how does one set the serverside settings? As there is no "settings" menu on a dedicated server.
Is there a config file that serveradmins can set like the old userconfig stuff? I googled only a bit on CBA settings but didnt find anything meaningful.

kavinsky:
ACE team does it with server settings.

dedmen:
Yep according to https://github.com/CBATeam/CBA_A3/tree/master/template/static_settings_addon
We just add a serverside only userconfig and some code to only load it on serverside. 
@zgmrvn would you try to add that to this PR please?

dedmen:
https://github.com/michail-nikolaev/task-force-arma-3-radio/commit/0819605973d8bb6830b109128b93b785fe3be749
I did the functions to load the settings file on preInit. Problem right now is CBA appears to dictate that all CBA_Settings.. settings have to be defined in userconfig\cba\settings.sqf. Because thats what https://github.com/CBATeam/CBA_A3/tree/master/optionals/auto_load_settings_file uses.
But this way we can't distribute the config file to not let people overwrite the CBA settings file.
But i think that's kinda fine. As only serveradmins will use this we might just add a wiki entry on how to set these settings (I don't know how the settings.sqf file should look like yet). And there are only few people that use this anyway.

@zgmrvn please also remove the old userconfig folder in the PR.

Need some opinions on what to do with the settings file @kavinsky @michail-nikolaev 
Im for just leaving it in userconfig\cba\settings.sqf and adding info to the wiki.

kavinsky:
#956

zgmrvn:
Looks like "Static settings" is something diferent than the "Settings Framework". This PR is fonctionnal and you should not have to code any additional function to load data.

CBA Settings has a menu you can access while being ingame in Options >> Game >> Configure Addons (at the bottom of the window).
http://images.akamai.steamusercontent.com/ugc/269474880029576673/1E54995EB12658C358B5A7DA5D91612582A50F5B/

You have to be logged as admin to access the server tab. Data are stored in client/server profile.

Here is the documentation i have used : https://github.com/CBATeam/CBA_A3/pull/317

dedmen:
Wow nice. I didnt know there is a Server tab there.
@kavinsky We should merge this then.

kavinsky:
Ill take it.

kavinsky:
the PR looks good for me, im gonna try to test it this night.

kavinsky:
@zgmrvn can you join the TFAR Slack room?

zgmrvn:
okay, zigomarvin@hotmail.fr

MakatoMiela:
It is recommended to execute the function via a CBA XEH preInit event. This way you can make sure that the setting is available in the ## Eden-Editor.

@https://github.com/CBATeam/CBA_A3/wiki/Settings-System
@https://github.com/CBATeam/CBA_A3/wiki/Extended-Event-Handlers-(new)#extended_preinit_eventhandlers

dedmen:
@MakatoMiela Already solved in 1.0. And i think it should also be in 0.9.12 already.. but.. maybe not

Issue #1098, opened 2016-08-18
TFAR not compatible with TS API 21


DerAtrox:
Hey, i just updated to the latest TeamSpeak version (TS 3.1), which comes with a new API version. 

Will TFAR support API version 21 in the future or will this project die with the new API?

FATKIDSLOV3CAK3:
Would really like to hear that this will be a up coming fix As the community i am apart of as a heavy use of TFR for our mod.

ArMaTeC:
would love to see this fixed also

psykzz:
+1

skapytek:
pls fix this.

dedmen:
see PR #1100 I'll add the finishing touches to that in the next few hours

FATKIDSLOV3CAK3:
Yh i am well aware of it being beta and not full release but i am the kinda of person to see what is coming in updates to come and it is best to try and get this done and or ready for when they push the full release of 3.1.

Better to have it done and ready then still be in the dark or not know about a problem :D

dedmen:
My PR for the update is ready you can compile it yourself or wait till @michail-nikolaev is back from being AFK and does a release.

skapytek:
How can i compile it to get it working with newest ts3 update?

skapytek:
How can i compile it to get it working with newest ts3 update?

skapytek:
How can i compile it to get it working with newest ts3 update?

dedmen:
Here is a compile of my pull-request for 3.1

https://mega.nz/#!Q0EiwJCa!CW6DXdFFCYnY-fVVAwb9YGy_X1uqVrzycbacxm_Neiw
Its compiled with Visual Studio 2015 instead of 2010 in the official builds. And only 64 Bit because im too lazy for more. Use on your own risk.

skapytek:
Hi,
its not working with the newest ts3 update (3.1)
API version is still 20 but it needs version 21

regards

dedmen:
@skapytek  i just tested in my virtual machine. Installed a fresh x64 3.1 client and tried to enable my plugin. worked just fine. Keep in mind its only the 64bit version.. so if you have 32bit installation and both plugins in your plugins directory and only replace the 64bit one it will still use the 32bit one thats not updated.

skapytek:
I reinstalled teamspeak 3.1.
I deleted the 32bit .dll.
It wasnt working

dedmen:
[...]

PR #1100, opened 2016-08-20
Update for Teamspeak version 3.1

dedmen:
This pull request is not that clean.. the change to getPluginPath actually breaks compatability for older API versions.. I can implement a workaround for that.. which would add about 4 more lines of code to that area. Tell me if you are fine with that. Other solution would be to just not support any older versions. Which i wouldnt like.. because.. I dont like updating Teamspeak because I keep finding exploits for it.. There is a new crash exploit for the 3.1 client already.. and its still in beta...

Notes for workaround:
grab getPluginPath as ptr and wrap into ptr for
void (*getPluginPath)(char* path, size_t maxLen);
if API version returned by ts3plugin_apiVersion() < 21


dedmen:
Added the hack to support <v21
I couldnt compile because im using MSVC2015 and my base of this pull-request was still on older version.
But none of the errors i had were in plugin.cpp so im guessing this is okey.

dedmen:
Btw @michail-nikolaev if you want to update to VS 2015 you need to remove the /Fx flag.

dedmen:
Just did a successful compile for #1098 but didnt test the plugin in Teamspeak 3.1 yet. But just assuming its working fine.

dedmen:
Needs testing

- [ ]  3.0.19.4 64bit
- [x] 3.0.19.4 32bit
- [ ] 3.1 64bit
- [x] 3.1 32bit

dedmen:
Merged to correct branch in #1118

PR #1103, opened 2016-08-23
Update plugin version for TS3 Beta

psykzz:


dedmen:
Doesnt work because you didnt include the changed API. #1100 is probably better although im not done with that yet

dedmen:
As this PR doesnt actually work because the changed API functions and classes are in no way respected I will close this.

PR #1112, opened 2016-09-16
Fixes #1109

### LINKED ISSUE #1109: Add NamedPipe transact timeout.
Now when the Teamspeak plugin doesnt answer a request from the pipe arma freezes. Which may happen if the Teamspeak plugin crashes.
TransactNamedPipe (https://msdn.microsoft.com/de-de/library/windows/desktop/aa365790(v=vs.85).aspx)
offers to use Overlapped IO which one can use to implement a timeout while checking if data is available for read yet. 
I may do a pull request for that but im too busy so ill just keep this here as a reminder.

dedmen:
#1109


dedmen:
Tested locally so merging to 1.0 branch.

PR #834, opened 2014-10-28
Radio On/Off

CorruptedHeart:
Currently there is no way to tell if a radio is on/off. When pressing the tangent, it might be better to `showHint` a message saying that the radio is off.

I will begin working on disabling a couple UI elements (channel and frequency) when opening the display when the radio is off. Power buttons will have to be added to the radio dialogs, or the volume rotator should stop at 0 and the radio should be placed in an off state.

Relies on https://github.com/michail-nikolaev/task-force-arma-3-radio/pull/833
Partially solves https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/753

Issue #753, opened 2014-10-12
Ability to switch radios on/off


CorruptedHeart:
Being able to switch radios on/off which would disable the ability to use them. If a radio is switched off, it can avoid being sent to the plugin and potentially reduce load (?). If we introduce a battery feature, this could be a way of countering the battery running out.

But essentially it will be setting the volume to 0 for the most part.

CorruptedHeart:
Closed via #1132

PR #1132, opened 2016-09-23
Radio on off

CorruptedHeart:
Currently there is no way to tell if a radio is on/off. When pressing the tangent, it might be better to `showHint` a message saying that the radio is off.

I will begin working on disabling a couple UI elements (channel and frequency) when opening the display when the radio is off. Power buttons will have to be added to the radio dialogs, or the volume rotator should stop at 0 and the radio should be placed in an off state.

Relies on https://github.com/michail-nikolaev/task-force-arma-3-radio/pull/833
Partially solves https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/753

Original PR: https://github.com/michail-nikolaev/task-force-arma-3-radio/pull/834


arma3/@task_force_radio/addons/task_force_radio/functions/fn_sendFrequencyInfo.sqf
```diff
@@ -55,6 +56,7 @@ if (((call TFAR_fnc_haveLRRadio) or (TFAR_currentUnit != player)) and {[TFAR_cur
 		_radios = _radios + (player call TFAR_fnc_lrRadiosList);
 	};
 	{
+    if ([_x] call TFAR_fnc_RadioOn)then{
```
dedmen: Closing '}' ?

Issue #1134, opened 2016-09-23
Fix TFAR_fnc_isAbleToUseRadio


dedmen:
isAbleToUseRadio currently returns if player is unable to use Radio.
Additionally that function is missing the description header.

PR #1135, opened 2016-09-24
Fix TFAR_fnc_isAbleToUseRadio

CorruptedHeart:
Fix for: #1134 
Also added headers to scripts that were changed as part of this fix.

Currently untested due to the build process being broken due to the various structural changes that have occured.


dedmen:
Not much to go wrong here. Thank you.

Issue #1123, opened 2016-09-20
Arsenal save causes duplicate radio ID's


dedmen:
zgmrvn [11:45 PM]  
let's say i save my gear with the radio TFAR_r7800_1

zgmrvn [11:46 PM]  
and in a fresh multiplayer game, when a player serlecte a radio, he recieve the TFAR_r7800_1, then i load my gear with the TFAR_r7800_1 too

dedmen [11:49 PM]  
ah i see.. @zgmrvn you have a saved loadout with radio _1 when you load that after another guy already got _1 you now have two _1 radios... Mh.... Dont know how one would fix that... Please create an issue for that

Solution ideas:

zgmrvn [11:54 PM]  
a DialogEventHandler would be better

kavinsky [11:55 PM]  
if we overwrite Arsenal save function we need to check every release they do

dedmen [11:56 PM]  
@zgmrvn please create a issue for that so we can keep notes there

dedmen [11:55 PM]  
if we can catch that a save has just been done.. We can manually iterate through saved loadouts in profileNamespace and change radios there
[11:57]  
But we can attach one to the save button and then do a spawn which then does our profileNamespace stuff

dedmen [11:58 PM]  
because the save is called by the button which will run in unscheduled .. when we spawn we get the next scheduled tick when the save will be done

[11:59]  
but.. that solves the arsenal problem.. not the possible Eden Loadout problem

[11:59]  
would have to check if loadouts in eden missions save with the radio id

kavinsky:
Reference: #313

zgmrvn:
`player addEventHandler ["Take", {}];` is not triggered by VA

dedmen:
We can just iterate through the saved loadouts in profilenamespace before
each missionstart. That would make sure all saved loadouts are clean before
mission start... And you cant get duplicate radios inside one mission if
you started with clean loadouts. You can create corrupt loadouts and load
them again while playing the mission. But that cant create a duplicate
radio id

zgmrvn:
so you could not have a radio in your saved gear ?

dedmen:
Radios yes. But not radios with IDs

dedmen:
But that would not fix the Eden Loadout issue.. If thats even existent.. I'll check that now

BryanBoru:
Although I use Arsenal more than VAS recently, I am excited to see this addressed.

dedmen:
Fix merged in #1139

PR #1139, opened 2016-09-26
Loadout cleaning to replace actual radios with prototypes

zgmrvn:
https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/1123


addons/core/functions/fnc_loadoutReplaceProcess.sqf
```diff
@@ -0,0 +1,49 @@
...
+	// iterate through each container
+	{
+		_content = _x;
+
+		// iterate through each item of the container
+		{
+			_class = ConfigFile >> "CfgWeapons" >> _x;
+
+			// if the item is actual radio, not a radio prototype
+			if ((isClass _class) && (isNumber (_class >> "tf_radio"))) then {
```
CorruptedHeart: You can use TFAR_fnc_isPrototypeRadio passing the class name to perform the same function.

zgmrvn: yeah but it doesn't make the difference between a common item and a radio_id class so i still have to check it after

addons/core/functions/fnc_ClientInit.sqf
```diff
@@ -1,5 +1,5 @@
-#include "script_component.hpp"
-
+#include "script_component.hpp"
+
```
dedmen: What is this? 😆

dedmen: okey its LF vs CRLF wow :D

dedmen:
Worked fine in my test. Couldn't see anything that could wrong here. Because we are only taking everything by reference in case of error only thing that could happen is nothing getting changed.

addons/core/functions/fnc_ClientInit.sqf
zgmrvn: it didn't wanted to reject this bloc -__-', lack of skill in git stuff

dedmen: dont even know why that block had LF instead of CRLF which is everywhere else.

addons/core/functions/fnc_loadoutReplaceProcess.sqf
```diff
@@ -0,0 +1,49 @@
...
+		_content = _x;
+
+		// iterate through each item of the container
+		{
+			_class = ConfigFile >> "CfgWeapons" >> _x;
+
+			// if the item is an actual radio, not a radio prototype nor common item
+			if ((isClass _class) && (isNumber (_class >> "tf_radio"))) then {
+				// find his parent prototype
+				_parent = ([_class, true] call BIS_fnc_returnParents) select 1;
```
dedmen: Please use tf_parent instead https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/1.0/addons/handhelds/pnr1000a/CfgWeapons.hpp#L13
Im planning to change the inheritance on instanced radio classes a bit. And tf_parent is just meant to be used for that.

zgmrvn: ok, will tf_parent be a number representing a boolean ? Being true (1) only if the class refers to a prototype radio ? Or present only in prototype radios ?

dedmen: tf_parent is a string of the classname of the parent
click my link ;)

zgmrvn: Ooops, sorry. Nice, I go for the changes later.

zgmrvn: but, if the parent has the property too, how do i know if it's a prototype or an actual radio ?

dedmen: Only use that to find whats the parent.. as a replacement of calling "BIS_fnc_returnParents select 1" because that will just return the first Parent. And in my planned changes one would have to use the second parent.

addons/core/functions/fnc_loadoutReplaceProcess.sqf
```diff
@@ -33,11 +33,8 @@ for [{private _i = (count _loadouts) - 1}, {_i > 0}, {_i = _i - 2}] do {
 
 			// if the item is an actual radio, not a radio prototype nor common item
 			if ((isClass _class) && (isNumber (_class >> "tf_radio"))) then {
```
CorruptedHeart: The second part of the if should be in `{ }` for lazy eval.
`if ((isClass _class) && {isNumber (_class >> "tf_radio")}) then {`

dedmen:
How much optimization one can get by just not merging a PR for a few days :D

zgmrvn:
^^ the holly BIS Code Optimization Page says it !

dedmen:
I'd like to rename your function. TFAR_fnc_loadoutReplaceProcess is too generic and one doesn't know whats behind that... 
i thought about TFAR_fnc_arsenalReplaceNonProtypeRadios but that seems a bit too long.. Do you maybe know something better?

zgmrvn:
hey, no problem, i re-used the code of this function yesterday and told myself the same. Hum, hard to find something as representative as TFAR_fnc_arsenalReplaceNonProtypeRadios :/

PR #1142, opened 2016-09-29
Hide freq

ArMaTeC:
add the ability to hide frequency to make it more streamer friendly


dedmen:
better move that hiding stuff to fnc_GetChannelFrequency  That way we have it only in one place and not copy pasted all over the place

but we would also have to add a parameter to getChannelFrequency and getXXFrequency and currentXXFrequency

but i really dont like the style of that copy-paste everywhere. so.. parameter or an extra function like
_freq = _freq call TFAR_fnc_doStreamFriendlyUIStuff

that checks if it should be censored or not

ArMaTeC:
do what you feel it best guys my way was just a hacky way

dedmen:
These changes can be made inline and need a CBA Setting anyway. This is too hacky for 1.0. I'm gonna redo that later. @ArMaTeC I'm gonna message you on Slack before i redo this. (Just writing this so I remember)

Issue #923, opened 2015-04-07
ACE3 integration


michail-nikolaev:
- ability to open radios in self-interation menu
- ability to use other person LR radio via 3D menu (also, should be possible without ACE3)
- replace MicroDAGR in TFAR with the ACE 3 alternative, if such exists
- **"Take radio"** – 3D action over a dead player

michail-nikolaev:
* headphones integration (not implemented yet)

CorruptedHeart:
TODO
- [ ] Open Radios via self-interaction
- [ ] Use another player's radio via interaction
- [ ] Use ACE 3's MicroDAGR (if ace 3 is loaded)
- [ ] Take radio on dead body interaction
- [ ] Headphones integration (I assume this will be for in vehicles?)

Knight1981:
The 2nd point would be awesome!
"ability to use other person LR radio via 3D menu (also, should be possible without ACE3)"

It would be really cool if the "actual" (e.g. teamleader) can use the radio of the operator etc. :)

BryanBoru:
+1 this! thanks!

SzwedzikPL:
"Open vehicle radio via interaction menu in vehicles" @CorruptedHeart hm?

PR #1144, opened 2016-10-06
Add support for ACE3

CorruptedHeart:
Handheld radios that the player has will show under a Radio action within the self interact menu.
Currently it only supports opening the radio interface.

This is currently WIP. I just wanted to get it here in case anyone else wanted to test it or jump in on it.

Working to solve: #923 
### TODO:
- [x] Open Radios via self-interaction
- [x] Radio icon
- [ ] Move Radios under Equipment
- [ ] Open vehicle radio via interact menu in vehicles
- [ ] "Take radio" on dead body interaction
- [x] Allow using ACE's MicroDAGR to program personal radios

radio in ace

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,4 @@
 *.psd filter=lfs diff=lfs merge=lfs -text
 *.png filter=lfs diff=lfs merge=lfs -text
 *.lib filter=lfs diff=lfs merge=lfs -text
+*.ai filter=lfs diff=lfs merge=lfs -text
--- a/addons/core/ACEActions.hpp
+++ b/addons/core/ACEActions.hpp
@@ -6,7 +6,7 @@ class CAManBase: Man {
             condition = "call TFAR_fnc_haveSWRadio";
             exceptions[] = {"isNotSwimming"};
             statement = "";
-            icon = "";
+            icon = QPATHTOF(ui\ACE_Interaction_Radio_Icon.paa);
             insertChildren = "[_player] call TFAR_fnc_addRadiosToACE";
         };
     };
--- a/addons/core/ui/ACE_Interaction_Radio_Icon.paa
+++ b/addons/core/ui/ACE_Interaction_Radio_Icon.paa
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:c52da656457b6c38b3e56198df89292d4243551b15ff8444c9ee912f8b149f80
+size 22016
--- a/extra/textures/UI/ACE_Interaction_Radio_Icon.ai
+++ b/extra/textures/UI/ACE_Interaction_Radio_Icon.ai
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:4727e8eb20257ad4efb421ebaba0654cbe818e32e714c747ecabfe5705094767
+size 360129


dedmen:
If you change indentation of CfgVehicles.hpp anyway, also change tabs -> spaces

commy2:
nice

dedmen:
T´he ACE MicroDagr compat is not tested.

addons/core/functions/fnc_haveProgrammator.sqf
```diff
@@ -1,21 +1,21 @@
...
+    Nothing
+
+    Returns:
+    BOOLEAN
+
+    Example:
+    _hasProgrammer = call TFAR_fnc_haveProgrammator;
+*/
+if (isNil {TFAR_currentUnit} || {isNull (TFAR_currentUnit)}) exitWith{false};
+"tf_microdagr" in (assignedItems TFAR_currentUnit) || {"ACE_microDAGR" in (assignedItems TFAR_currentUnit)};
```
CorruptedHeart: For the compat with MicroDagr, I'd almost want to move that to a config option rather than hardcoded like this.

dedmen: Why would people want to disable that? But do as you wish ^^

CorruptedHeart: It's so we won't have to edit this file if we want to add more items as programmers.

dedmen:
The SW Radio submenu has an option to copy settings from one Radio to another. You select the Radio then "copy Settings From" and then you get a list of all Other Radios you have and you can select one to copy settings

dedmen:
Allow to attach Radio's to vehicles and stuff like you can do with chemlights and explosives.. That way one could also physically wear the Radio on the body.. but thats overkill.

dedmen:
Lower headset and hotkeys for lowering/raising headset

SageNTitled:
I would love to see a keybind for "lower headset" as well, for even quicker access. Doesn't have to be bound to anything, just an option to would be nice.

dedmen:
New problem concerning Attachable Radios. You can't just Attach pure Items.. You have to have a CfgVehicles class for each Attachable Item.. That means we need to have a Vehicle for each Radio.. Meaning 1000 additional vehicle classes per Radio.
Also ACE_Attach doesn't support assigned Items. Meaning it would only be attachable if the Radio is in the Inventory. Not in the Radio Slot. Ofcause that could theoretically be changed on ACE side.
But I am really not sure if we want another 6000 Vehicle classes (1000 for each Radio class) just for such a kinda small feature.

kavinsky:
@dedmen what about attaching a weaponholder?

dedmen:
Not possible with ACE Attach natively. Also Weapon holder is a bad Idea... because you can open the Weapon Holder and put more items into that Holder. @BaerMitUmlaut

dedmen:
Similear to allowing players to use other Peoples Radios allow them to use Vehicle Radios for example from a Cargo seat.

dedmen:
Add serverside option to allow using Remote Radios. (Default disabled probably.. As We may do harm if anyone can now access other People's Radios and change settings...)
Add Self action to set an inactive Radio to active.

commy2:
>Also Weapon holder is a bad Idea... because you can open the Weapon Holder and put more items into that Holder.

There is this idea to make weapon holders that only have space for one item and / or have zero / negative cargo range.

654wak654:
If you have a weaponholder with `isGround = 0;` in it's config, you get a transparent cargo holder with only the 'inventory' option, which can be prevented.
Also I used `supplyRadius = 0.001;` in the past to make it really hard to reach weaponholder actions.

BmanDesignsCanada:
Is there any way to work on this code? Can't find a branch or anything to fork.

kripto202:
I don't think that the take radio interact button is needed since you can easily get it through the inventory.

dedmen:
@kripto202 You can't as long as player is alive. And thus ACE Medical unconciousness.

kripto202:
@dedmen are you sure? There is an ace option for open inventory and you can pull the radio out there. I know you can if you have him surrender or have him tied up.

PauleyT:
Is it possible to make a ace self interaction to turn off/on the radio?

dedmen:
@PauleyT Yes. Using the new Radio On/Off feature. Good Idea actually.

dedmen:
Take Radio of dead body,
radio on/off,
remotely use others LR radio or vehicle radio

Are all now scratched off the 1.0 list so we can finally get a release done. Feel free to create new feature requests for the things that don't have them yet.

dedmen:
Thanks to @CorruptedHeart deleting his fork this is unmergeable :D 
#1315

CorruptedHeart:
Yeah, that was a my bad. Was cleaning up my repositories.

PR #1146, opened 2016-10-07
Change to use new private syntax and other minor improvements

nikolauska:
This merge request mainly change private to new syntax where private is added with variable and not separately.  There are also some whitespace fixes and other minor improvements.


dedmen:
Why is using private _variable better than private ["_variable1","_variable2"] ?

addons/core/functions/fnc_isSpeaking.sqf
```diff
@@ -1,5 +1,28 @@
 #include "script_component.hpp"
 
+<<<<<<< HEAD
```
dedmen: nope!

nikolauska: This has now been fixed, somehow managed to miss it

nikolauska:
Having private before variable allows more easily to see scope it can be used in and where that variable has been initialized. Defining it as an array on top of the file means it can be used on whole file which is not really wanted behavior as you want to limit variable to only scope they are used.

Also problem that I saw in private ARRAY usage was that people tend to forget to update it when changing code where as having it on variable initialization makes it easier to remember.

Of course this it can look stupid in some of the files currently where there are dozens of variables defined, but I wanted to avoid changing code too much in this pull request and focus on just this.

commy2:
`private keyword` is faster than `private command` (`private ARRAY` or `private STRING`).

If you compare older BI scripts like the Arsenal with newer ones like 3DEN, you can see that they use this newer syntax excusively now too.

nikolauska:
I remembered there was talk about it, but couldn't find any info about it so I didn't mention it

addons/core/functions/fnc_generateSwSettings.sqf
```diff
@@ -52,3 +52,4 @@ if (!_set) then {
 _sw_settings set [2, _sw_frequencies];
 
 _sw_settings
+>>>>>>> 8b8749952bfc75733054c34ae530f0cfdbf1d5bb
```
dedmen: möp! Also line 35 we are using 4 space tab. I'll fix that

dedmen:
Do a playtest please. you can get the build from appveyor

nikolauska:
There were two errors found on fnc_radioOn and fnc_onDDTangentReleasedHack on testing which have now been fixed.

addons/core/functions/fnc_onDDTangentReleasedHack.sqf
```diff
@@ -1,6 +1,6 @@
 #include "script_component.hpp"
 
-params ["_scancode"];
+params ["", "_scancode"];
```
dedmen: could use param instead of params to only get the second parameter

dedmen:
So. When you are sure you are done and everything is fine I will merge this. Better look over everything again to make sure it works fine. I did a new test build that you can test if needed.

addons/core/functions/fnc_currentDirection.sqf
```diff
@@ -18,22 +18,19 @@
     Example:
         call TFAR_fnc_currentDirection
 */
-private ["_current_look_at", "_current_hyp_horizontal", "_current_rotation_horizontal"];
 
-_current_look_at = (screenToWorld [0.5,0.5]) vectorDiff (eyepos TFAR_currentUnit);
-_current_look_at params ["_current_look_at_x", "_current_look_at_y", "_current_look_at_z"];
+private _current_look_at = (screenToWorld [0.5,0.5]) vectorDiff (eyepos TFAR_currentUnit);
+private _current_look_at params ["_current_look_at_x", "_current_look_at_y", "_current_look_at_z"];
```
dedmen: whats this? I thought you tested this. This spammed my RPT in my test. 
-rep

nikolauska: I did test this but my RPT didn't show any issues, sorry for missing that :(

PR #1198, opened 2017-01-30
Korean patch

sbng-PAPA:


dedmen:
This is for the Old TFAR files version 0.9.x that version is discontinued all new work is happening on 1.0
https://github.com/michail-nikolaev/task-force-arma-3-radio/tree/1.0
I could just move your changes to the new files but I don't know when I will have time for that.

PR #1210, opened 2017-03-01
change bw mr3000 encription code to west

commy2:


dedmen:
Wrong branch. 1.0 > master.
I'll fix it.

commy2:
>What was the old radio code? east?! wtf :D

It inherited from the other mr3000's and those are on east side.

>Wrong branch. 1.0 > master.

Is this correct now? I thought I targeted `master`.

dedmen:
:D That's the Problem. master is wrong ^^ 
You did target master and I retargeted to 1.0 but couldn't fix conflicts because I didn't have permission to push to your patch-1 branch

PR #1229, opened 2017-03-25
French Readme

11RDP-LoupVert:
Translated Readme, matching on website's [readme] (http://radio.task-force.ru/en/).


dedmen:
The stuff talking about Pipe errors is outdated. Also If you hadn't noticed we removed all non-english Readme's. We only plan to have English from now on we aren't enough native Speakers in the Team to always keep such stuff updated.

11RDP-LoupVert:
Roger Dedmen

PR #1247, opened 2017-04-18
Update script because cba update

shukari:
https://github.com/CBATeam/CBA_A3/pull/590


dedmen:
Thanks.

dedmen:
Btw this caused a script error because you need to
`([GVAR(radioTowerList)] call CBA_fnc_hashKeys)`

PR #1249, opened 2017-05-02
Small German Language Corrections

KiritoKun223:
- Change in STR_voice_yelling
  Just an idea: in english it's yelling, and in German it was "loud". Yelling is translated in "Schreien", so that might be more "correct".
- Change in STR_voice_whispering
  same as above
- Change in STR_init
  "initialization" is translated into "Initialisierung", just an idea as above. Not a mistake.
- Change in STR_radio_diver_text
  translation of "isolated" was missing. Not needed tho, but more correct
- Change in STR_radio_same_sw_frequencies_for_side
  upper case letters are used in that correlation. Can be seen in another example here: http://emf3.bundesnetzagentur.de/pdf/UWB-BNetzA.pdf
- Change in STR_radio_same_lr_frequencies_for_side
  same as above
- Change in STR_radio_same_dd_frequencies_for_side
  same as above

### DIFF
--- a/arma3/@task_force_radio/addons/task_force_radio/stringtable.xml
+++ b/arma3/@task_force_radio/addons/task_force_radio/stringtable.xml
@@ -1666,7 +1666,7 @@
 					Gritar
 				</Spanish>
 				<German>
-					Laut
+					Schreien
 				</German>
 				<Turkish>
 					Bağırma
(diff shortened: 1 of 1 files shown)


dedmen:
Wrong branch. new stuff should go to 1.0 branch.

KiritoKun223:
ah god damnit, allright

dedmen:
You also have moved the PR over to the new branch without closing it :/ I hate duplicate stuff ^^ Well.. too late

PR #1250, opened 2017-05-02
Some small corrections in German translation - Now in the right branc…

KiritoKun223:
…h! :D

- Change in STR_voice_yelling
  used translation of the word "yelling" and not "loud"
- Change in STR_voice_whispering
  same as above
- Change in STR_init
  used a noun like in english
- Change in STR_radio_same_dd_frequencies_for_side
  used upper case letters, can be seen here in another example: http://emf3.bundesnetzagentur.de/pdf/UWB-BNetzA.pdf
- Change in STR_radio_same_lr_frequencies_for_side
  same as above
- Change in STR_radio_same_sw_frequencies_for_side
  same as above
- Change in STR_radio_diver_text
  added "isolated" to German translation. Not needed, but more precise
- Change in STR_radio_control
  (not sure) should be "Steuerung" as in controlling (steuern)

### DIFF
--- a/addons/core/stringtable.xml
+++ b/addons/core/stringtable.xml
@@ -688,7 +688,7 @@
                 <Original>yelling</Original>
                 <Czech>křičení</Czech>
                 <French>Forte</French>
-                <German>Laut</German>
+                <German>Schreien</German>
                 <Italian>Gridare</Italian>
                 <Polish>krzyk</Polish>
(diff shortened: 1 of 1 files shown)


dedmen:
IMO `Initializing` and `Initialisieren` would be better.. But it's `Initialization` in english already so.. Whatever.

PR #1269, opened 2017-05-30
Ability to override half-duplex mode for selected radios

MorpheusXAUT:
As discussed in Discord, our community would love a feature that allows us to override the half-duplex restrictions for certain radios only.

Example:
We'd like for our infantry to use "normal" half-duplex radios, however our pilots could use full-duplex LR radios so they can continuously talk within their group and still notice a radio message by their assigned JTAC.

This PR implements an additional setting for SW and LR radios (`tf_halfDuplexOverride`), which allows mission designers to disable the restrictions for certain radios only. The TeamSpeak plugin and `sendFrequencyInfo` have been modified to include an additional flag, indicating whether the override is enabled or not.

I'm more than happy to work on any changes required to have this feature included or change the naming of this.

EDIT: I've tested and verified this addition with a few players, however a full mission based on this modification is still in the works and thus outstanding.


addons/backpacks/anarc164/CfgVehicles.hpp
```diff
@@ -14,5 +14,6 @@ class TFAR_anarc164: TFAR_Bag_Base {
     tf_dialog = "anarc164_radio_dialog";
     tf_subtype = "airborne";
     tf_dialogUpdate = "[""%1""] call TFAR_fnc_updateLRDialogToChannel;";
+    tf_halfDuplexOverride = 0; //0 disables half-duplex override, everything greater than 0 enables it
```
dedmen: Default to 0 should probably be inside TFAR_Bag_Base because that's also what 3rd party mods would inherit

dedmen: You can also have it in base&subclass doesn't matter much if it's duplicate

MorpheusXAUT: Alright, I'll push a change to add it to `TFAR_Bag_Base` as well once I'm home. I assume you mean adding it [here](https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/1.0/addons/backpacks/CfgVehicles.hpp#L10)?

dedmen: Yes. Thanks.

addons/core/XEH_PREP.sqf
```diff
@@ -65,6 +65,8 @@ PREP(getDefaultRadioClasses);
 PREP(getRadioOwner);
 PREP(getSwSpeakers);
 PREP(getLrSpeakers);
+PREP(getSwHalfDuplexOverride);
```
dedmen: New wording is SR(Short-Range) instead of SW (short-wave) and LR (Long-range)
Old function names were kept as Sw for backwards compat. But everything Internal that is not exposed as part of the API is using SR already. 
Might be confusing for people though that some things are Sw and some Sr.
This change was meant to make it easier for Players. Modders can probably cope with Sw meaning short-range.
Your choice.

addons/core/XEH_PREP.sqf
```diff
@@ -177,6 +179,8 @@ PREP_SUB(hint,showRadioVolume);
 PREP(setChannelFrequency);
 PREP(setRadioOwner);
 PREP(setLrSpeakers);
+PREP(setSwHalfDuplexOverride);
+PREP(setLrHalfDuplexOverride);
```
dedmen: You can probably put both into one function. LR Radio is always an Array and SR Radio is always a string. That way you can detect inside the function what was passed.
I think having it in one function is easier to use.

addons/core/functions/fnc_getLrHalfDuplexOverride.sqf
```diff
@@ -0,0 +1,24 @@
...
+            0: OBJECT - Radio object
+            1: STRING - Radio ID
+
+    Returns:
+        BOOLEAN - Half-duplex mode should be skipped for this SW radio
+
+    Example:
+        (call TFAR_fnc_activeLrRadio) call TFAR_fnc_getLrHalfDuplexOverride;
+*/
+
```
dedmen: Please use `params` to validate Input. Array for LR and String for SR Radios.

addons/core/functions/fnc_getLrSettings.sqf
```diff
@@ -86,6 +86,18 @@ if (isNil "_radioCode") then {
     _value set [TFAR_CODE_OFFSET, _radioCode];
 };
 
+private _halfDuplexOverride = _value select HALFDUPLEX_OVERRIDE_OFFSET;
```
dedmen: param with default value is IMO better then a isNil check.

addons/core/functions/fnc_getLrSettings.sqf
```diff
@@ -86,6 +86,18 @@ if (isNil "_radioCode") then {
...
+if (isNil "_halfDuplexOverride") then {
+    _halfDuplexOverride = false; //Override is disabled by default. No harm done if we don't find the config value since it's expected default behaviour anyways
+    private _override = [_radio_object, "tf_halfDuplexOverride"] call TFAR_fnc_getLrRadioProperty;
```
dedmen: doesn't `getLrRadioProperty` also have a parameter for default value that you can use?

addons/core/functions/fnc_getLrHalfDuplexOverride.sqf
dedmen: Oh. I forgot. That would break then if you call it just with an array. hmpf.
Maybe a `isEqualType []` check just to be safe? It's probably not worth the effort.

addons/core/XEH_PREP.sqf
MorpheusXAUT: Fair enough, I was getting a bit confused by that as well. Will change it to SR.

addons/core/functions/fnc_setSwHalfDuplexOverride.sqf
```diff
@@ -0,0 +1,27 @@
...
+        1: BOOLEAN - Half-duplex override enabled
+
+    Returns:
+        Nothing
+
+    Example:
+        [call TFAR_fnc_activeSwRadio, true] call TFAR_fnc_setSwHalfDuplexOverride;
+*/
+
+params ["_radio_id", "_halfDuplexOverride"];
```
dedmen: type checking inside params

ts/src/clientData.cpp
```diff
@@ -150,7 +150,8 @@ LISTED_INFO clientData::isOverLocalRadio(std::shared_ptr<clientData>& myData, bo
         LockGuard_shared lock(&TFAR::getInstance().m_gameData.m_lock);
         auto &frequencyInfo = TFAR::getInstance().m_gameData.myLrFrequencies[senderFrequency];
         if (!TFAR::config.get<bool>(Setting::full_duplex) && //If we are currently transmitting on that Radio we can't hear so we return before result gets valid
-            frequencyInfo.radioClassname.compare(currentTransmittingRadio) == 0) {
+            frequencyInfo.radioClassname.compare(currentTransmittingRadio) == 0 &&
+			!frequencyInfo.halfDuplexOverride) { //Certain radios can be capable of using full-duplex as an exception, check for half-duplex override
```
dedmen: Check this first before the classname check. A bool check is faster than a string compare.
Should probably move this into a function. But that's my Job not yours ^^

addons/core/XEH_PREP.sqf
MorpheusXAUT: Sounds like a good idea, I'll merge both functions and rename it to `setHalfDuplexOverride`/`getHalfDuplexOverride`. Will just have to add a few checks for array/string to distinguish, but shouldn't be a problem since the functions don't actually include much logic anyways.

ts/src/clientData.cpp
```diff
@@ -160,7 +161,8 @@ LISTED_INFO clientData::isOverLocalRadio(std::shared_ptr<clientData>& myData, bo
         LockGuard_shared lock(&TFAR::getInstance().m_gameData.m_lock);
         auto &frequencyInfo = TFAR::getInstance().m_gameData.mySwFrequencies[senderFrequency];
         if (!TFAR::config.get<bool>(Setting::full_duplex) && //If we are currently transmitting on that Radio we can't hear so we return before result gets valid
-            frequencyInfo.radioClassname.compare(currentTransmittingRadio) == 0) {
+            frequencyInfo.radioClassname.compare(currentTransmittingRadio) == 0 &&
+			!frequencyInfo.halfDuplexOverride) { //Certain radios can be capable of using full-duplex as an exception, check for half-duplex override) {
```
dedmen: same. 
```
!frequencyInfo.halfDuplexOverride &&
frequencyInfo.radioClassname.compare(currentTransmittingRadio) == 0
```

ts/src/helpers.cpp
```diff
@@ -279,12 +279,17 @@ std::map<std::string, FREQ_SETTINGS> helpers::parseFrequencies(const std::string
     std::vector<std::string> v = split(sub, ',');
     for (const std::string& xs : v) {
         std::vector<std::string> parts = split(xs.substr(1, xs.length() - 2), '|');
-        if (parts.size() == 3 || parts.size() == 4) {
+        if (parts.size() == 3 || parts.size() == 4 || parts.size() == 5) {
```
dedmen: Can parts still be 4 elements? classname and override are both part of half duplex. So 4 elements shouldn't be possible anymore,

ts/src/task_force_radio.cpp
```diff
@@ -111,19 +111,22 @@ TFAR::TFAR() {
         diag << TS_INDENT << TS_INDENT << "pttDelay: " << m_gameData.pttDelay.count() << "ms\n";
 
         std::array<char*, 3> stereoModeToString{ "stereo","left","right" };
+		std::array<char*, 2> halfDuplexOverrideToString{ "disabled", "enabled" };
```
dedmen: Wrong Indentation. you are using tabs here. Same goes for helpers.cpp

ts/src/task_force_radio.cpp
```diff
@@ -111,19 +111,22 @@ TFAR::TFAR() {
...
         diag << TS_INDENT << TS_INDENT << "mySRFreq:\n";
         for (auto& it : m_gameData.mySwFrequencies) {
             diag << TS_INDENT << TS_INDENT << TS_INDENT << it.first << ":\n";
             diag << TS_INDENT << TS_INDENT << TS_INDENT << TS_INDENT << "radio: " << it.second.radioClassname << "\n";
             diag << TS_INDENT << TS_INDENT << TS_INDENT << TS_INDENT << "stereo: " << stereoModeToString[static_cast<uint8_t>(it.second.stereoMode)] << "\n";
             diag << TS_INDENT << TS_INDENT << TS_INDENT << TS_INDENT << "vol: " << it.second.volume << "\n";
+			diag << TS_INDENT << TS_INDENT << TS_INDENT << TS_INDENT << "halfDuplexOverride: " << halfDuplexOverrideToString[it.second.halfDuplexOverride] << "\n";
```
dedmen: Does that compile? I think you also need to cast to uint8_t as with stereoModeString.

ts/src/task_force_radio.cpp
dedmen: and task_force_radio.hpp. Probably all plugin code you wrote is using tabs.

addons/core/functions/fnc_getLrSettings.sqf
MorpheusXAUT: Not that I am aware of, but I might also be reading something wrong... [The current version](https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/1.0/addons/core/functions/fnc_getLrRadioProperty.sqf#L22) only lists two params though, `_radio` and `_property`.

addons/core/functions/fnc_getSwSettings.sqf
```diff
@@ -65,6 +65,20 @@ if (isNil "_rc") then {
     _value set [TFAR_CODE_OFFSET, _rc];
 };
 
+private _hdo = _value select HALFDUPLEX_OVERRIDE_OFFSET;
```
dedmen: `_hdo` ? 
Readability doesn't hurt. Also I'd like this code to look as similear as possible to the LR code.
so `param` with default value.
And below you use isNumber inst.... Wait. Yeah no. Don't use direct config Access. Use our getConfigValue or what it's called. Our function provides integrated caching.

ts/src/task_force_radio.cpp
MorpheusXAUT: It did compile and print fine for me. I'll check again to make sure when I reformat the files though.

addons/core/defines.hpp
```diff
@@ -13,6 +13,7 @@
 #define RADIO_OWNER 7
 #define SPEAKER_OFFSET 8
 #define POWER_OFFSET 9
+#define HALFDUPLEX_OVERRIDE_OFFSET 10
```
dedmen: This is a serious API change. Some other people store Data at the End of that array. Even though they shouldn't.
Need to specify clearly that the first 30 elements of that Array are reserved for TFAR use.
#TODO for me.

addons/core/XEH_PREP.sqf
dedmen: Be careful about https://github.com/michail-nikolaev/task-force-arma-3-radio/pull/1269#discussion_r125027361 in the get Variant.

addons/core/functions/fnc_getSwSettings.sqf
MorpheusXAUT: Agreed on `_hbo`, that could have definitely been named better.
`TFAR_fnc_getConfigProperty` unfortunately won't work since [it only works for `CfgVehicles`](https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/1.0/addons/core/functions/fnc_getConfigProperty.sqf#L12), while a SR-Radio's config is stored in `CfgWeapons`. I think I was originally trying to use `TFAR_fnc_getConfigProperty`, but ran into this very issue and decided to deal with it the same way the rest of this file does.

addons/core/defines.hpp
MorpheusXAUT: Happy to change it to a different offset, I was already figuring this would be a problem. Does ArmA just allow one to access 0-9 and suddenly e.g. 30 (with possibly nothing inbetween)?

Or is there another way for me to properly store the setting?

addons/core/functions/fnc_getSwSettings.sqf
dedmen: There has to be something for SR Radios. The Range and other stuff is also read from config.

dedmen: https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/7c2fc03dcf659c54f0b12a1c4fc98b054b40f4bc/arma3/%40task_force_radio/addons/task_force_radio/functions/fn_onSwTangentReleased.sqf#L8
Wow. Okey.
Leave it then. I'll have to add a function for CfgWeapons later on.

addons/core/XEH_PREP.sqf
MorpheusXAUT: Yeah, I'll see if I can make it work in one function. If it gets too much work or hassle of checking which type we're handling, I might just leave it separated.
Will try to merge it into one function though!

addons/core/defines.hpp
dedmen: You are doing everything correct here. But last time we added an element to that Array some Mods had problems. Just need to make it public that the first 30 are reserved for us.

MorpheusXAUT: Alright. I'll keep it as it is then.

addons/core/functions/fnc_getSwSettings.sqf
MorpheusXAUT: Yeah, I looked around the source to find an alternative to direct config access, since I was trying to avoid it as well. My search was - same as yours - not very successful either...

Will stay as it is for now, until a `getConfigProperty` for `CfgWeapons` has been added. Happy to add another PR and change it at this point then (if not already done).

ts/src/task_force_radio.cpp
MorpheusXAUT: I'll check my VS settings and make sure it uses spaces, sorry about that. Will reformat properly.

ts/src/helpers.cpp
MorpheusXAUT: I mainly kept it for backwards compatibility, so a newer TS plugin version would still be able to parse an older ArmA addon command. Removing the check for `parts.size() == 4` would break the frequency modification for TS if the server is running an outdated version.

I'm happy to remove the additional check if you don't think this backward compatibility is important enough to maintain it.

addons/core/functions/fnc_getLrHalfDuplexOverride.sqf
MorpheusXAUT: Having a `params` check that would allow for both SR and LR types would probably end up being something like `params [["_radio", "", [["", ""], ""]]]`, which would have to be checked for `array`/`string` to figure out whether it's LR or SR, right?

addons/core/functions/fnc_getLrSettings.sqf
MorpheusXAUT: Will do.

addons/core/functions/fnc_setSwHalfDuplexOverride.sqf
MorpheusXAUT: Will do.

ts/src/clientData.cpp
MorpheusXAUT: Very fair point, I didn't think about optimization and speed as much as you probably have so far :smile:
Will change it around.

addons/core/functions/fnc_getLrHalfDuplexOverride.sqf
dedmen: Yeah. but the problem is if you pass an array to params and params only accepts type ARRAY.
It will check if the first entry in that Array is the datatype it wants.
So instead of 
`(call TFAR_fnc_activeLrRadio) call TFAR_fnc_getLrHalfDuplexOverride;`
You would need to
`[(call TFAR_fnc_activeLrRadio)] call TFAR_fnc_getLrHalfDuplexOverride;` Which is ugly.

If you have more parameters it's fine because you need to
`[(call TFAR_fnc_activeLrRadio), moreParameters] call TFAR_fnc_getLrHalfDuplexOverride;`
anyway.
So you essentially can't use params for functions that get the LR Radio directly.

MorpheusXAUT: Ah, I see... Yeah, that does indeed look ugly for the get-functions...
Would an alternative be skipping the `params` call and doing manual "verification" (by e.g. checking `isEqualType ""` for SR and `isEqualType []` for LR)?

addons/core/functions/fnc_getLrSettings.sqf
MorpheusXAUT: Looking at this code again, I think the `isNil` check is actually appropriate here.

Instead of only trying to get a default value (which would indeed be more elegant using `param` with a defaultValue), this code actually checks whether the setting is **missing** and adds it as appropriate.

Were I to use `private _halfDuplexOverride = _value param [HALFDUPLEX_OVERRIDE_OFFSET, false];`, I would have no way to determine whether the value was missing from this specific LR radio or whether it was set to `false` already. Thus, I could not reliably figure out whether to actually check the config of the LR radio for the value defined there without risking overwriting a setting for a specific radio with the default value of its class.

I hope this makes any sense to you, writing it was absolutely confusing to me already 😄

addons/core/functions/fnc_getLrHalfDuplexOverride.sqf
dedmen: Yeah. You could then just diag_log or use CBA's WARNING macro. 
A example of the warning macro can be found on the serverside Init stuff. For the deprecated Variables

MorpheusXAUT:
Most changes mentioned above have been added. The merging of SR/LR functionality into one script is still missing (and with that the param validation mentioned here https://github.com/michail-nikolaev/task-force-arma-3-radio/pull/1269#discussion_r125082074)

addons/core/functions/fnc_getLrHalfDuplexOverride.sqf
MorpheusXAUT: Alright, will do. As mentioned below, most stuff has already been fixed and pushed, I'll fix this with the merge of SR/LR functions next week.

addons/core/functions/fnc_getLrSettings.sqf
dedmen: Yeah. You could also use stuff like scriptNull as default value and then do a isNull check. I did that somewhere else already.
But doesn't matter then.

addons/core/functions/fnc_getLrSettings.sqf
dedmen: https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/1.0/addons/core/functions/fnc_getConfigProperty.sqf#L17 has a default value.
But that's not exposed through `getLrRadioProperty`. That's on my Todo list now then.

ts/src/helpers.cpp
dedmen: We are in Beta phase right now. We don't need to care about being compatible to older beta builds.
I regularly break backwards compat.

ts/src/task_force_radio.cpp
dedmen: If it works then it's probably fine.
Never tried to use array[bool]. But I guess the implicit integer conversion kicks in here.

MorpheusXAUT: Yeah, it implicitly gets casted to an int anyways, so works fine. You're right though, a cast to `uint8_t` makes it a bit more clear.

ts/src/helpers.cpp
MorpheusXAUT: If that's the case, couldn't we drop `parts.size() == 3` as well? Or is there anything else beside `TFAR_fnc_sendFrequencyInfo` triggering the frequency parsing? A quick search did not releave any special results, but I might have overlooked something.

dedmen:
Repository will get a full history rewrite which will nuke all PR's
If you still need this please create a issue as feature request.

PR #1273, opened 2017-06-09
Fixed bugs

gandjustas:
Two annoying bugs:
1) For games without respawn. If player killed during radio transmission  then transmission continues in teamspeak. Only way to stop it - restart TS. Added RELEASE_ALL_TANGENTS jn killed event.
2) For big (150+ players) games without respawn. Often players kicked by battle eye after mission start. After reconnect player apper without radio.

No way to fix it "localy". Many game servers uses TFAR, if I make fix for one of them, I'll need to use different version for other servers.

Same issues presents in 1.0 branch.

### DIFF
--- a/arma3/@task_force_radio/addons/task_force_radio/functions/fn_radioReplaceProcess.sqf
+++ b/arma3/@task_force_radio/addons/task_force_radio/functions/fn_radioReplaceProcess.sqf
@@ -48,6 +48,8 @@ while {true} do {
 					};
 					true;
 				} count _items;
+				//Bug fix for sectrator mode, player continue talk after being killed
+				"task_force_radio_pipe" callExtension (format ["RELEASE_ALL_TANGENTS	%1", name _unit]);
 			}];
 			TFAR_currentUnit setVariable ["tf_handlers_set", true];
 		};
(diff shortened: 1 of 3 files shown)


dedmen:
What are you trying to do here? I already told you we don't accept PR's on master branch. And you just ignored my further comments about problems in your changes and just reopened the same Pullrequest although I already denied it?
Also If you want to notify Teamspeak that a player has been killed. Why don't you put it into https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/master/arma3/%40task_force_radio/addons/task_force_radio/functions/fn_sendPlayerKilled.sqf Which does exactly what you want?
Also a Player really shouldn't be able to Talk when dead. That should be handled in the Teamspeak plugin on the Killed event. And not in SQF.

gandjustas:
I play with TFAR on 4 different server\projects, they are all:
1) 120+ players. Tomorrow we will play game with 210 players.
2) Don't have respawns, but have spectator mode, where all killed players can talk.
3) Use same TFAR version with signed PBOs. I can't push my fixes to all servers (it will take too long time), and it's hard to have different versions of TFAR for different servers.
4) Have own addons based on TFAR.

You can accept PR (may be hotfix branch) and create release in 10 minutes. All servers will get new version in couple days. That's what I trying to do.

PR #1309, opened 2017-09-13
Fix the loss of radio

MrDell61:
If the player inventory is full, then when change the radio, the old one is deleted


dedmen:
We don't currently accept PR's to master as 1.0 changed a lot of stuff. Also your fix is kinda crude.. I have a better Idea I think.
Use https://community.bistudio.com/wiki/canAdd before unassignItem to know beforehand if unassignItem will fail to move the Item into inventory.

PR #1310, opened 2017-09-13
Fix the loss of radio v2

MrDell61:
If the player inventory is full, then when change the radio, the old one is deleted. So most likely easier.

### DIFF
--- a/addons/core/functions/fnc_setActiveSwRadio.sqf
+++ b/addons/core/functions/fnc_setActiveSwRadio.sqf
@@ -21,4 +21,8 @@
 private _old = (call TFAR_fnc_activeSwRadio);
 TFAR_currentUnit unassignItem _old;
 TFAR_currentUnit assignItem _this;
+_listRadios = TFAR_currentUnit call TFAR_fnc_radiosList;
+if !(_old in _listRadios) then {
+	TFAR_currentUnit addItem _old;
+};
 ["OnSWChange", [TFAR_currentUnit, _this, _old]] call TFAR_fnc_fireEventHandlers;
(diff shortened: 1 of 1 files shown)


dedmen:
afaik assignItem won't do anything if the slot is already occupied.
What I was thinking of was 
```
private _couldAdd = canAdd _old;
TFAR_currentUnit unassignItem _old;
TFAR_currentUnit assignItem _this;
if (!_couldAdd) then {TFAR_currentUnit addItem _old;}; //We couldn't put it into inventory before but now we have space for sure.
```

MrDell61:
No, If there is no space in inventory the item simply disappears. https://community.bistudio.com/wiki/unassignItem
But if you want, you can do as you propose.

dedmen:
But in your example you were always adding the item.. even if unassign item might have succeeded. Causing duplicates.

MrDell61:
I did as you asked.

dedmen:
Looks correct for you too? right?

MrDell61:
With this use, they work the same way.

dedmen:
https://community.bistudio.com/wiki/canAdd syntax was wrong. object was missing

MrDell61:
Sorry, I did not check

shukari:
is not https://community.bistudio.com/wiki/linkItem better?

MrDell61:
> is not https://community.bistudio.com/wiki/linkItem better?

> But in your example you were always adding the item.. even if unassign item might have succeeded. Causing duplicates.

dedmen:
No. Because the item is moved from Inventory to the slot. linkItem would create a duplicate.

PR #1323, opened 2017-10-23
ACE3-Action: Stereo, inside Vehicles

### LINKED ISSUE #1328: ACE³ self interactions for interaction with radios not available if sitting in vehicle
**Arma 3 Version:** `current` (stable)
**CBA Version:** `3.4.1.170912`
**Task Force Radio Version:** `1.0.264`
**ACE3 Version:** `3.11.0`

**Mods:**
- `@CBA_A3`
- `@task_force_radio`
- `@ace`

**Description:**
- The self interactions "Radios" and "Lower Headset"/"Raise Headset" aren't shown when sitting in vehicle.

**Steps to reproduce:**
- Get in vehicle
- Try to lower your headset with ace³ self interaction.

Dorbedo:
* the stereo settings added to the self interaction menu
* fix for the default value in TFAR_fnc_getSwStereo
* add the possibility to interact from inside vehicles, on chairs
* fixes #1328

### DIFF
--- a/addons/core/ACEActions.hpp
+++ b/addons/core/ACEActions.hpp
@@ -4,8 +4,8 @@ class CAManBase: Man {
         class ACE_Equipment {
             class TFAR_Radio {
                 displayName = "Radios";
-                condition = "call TFAR_fnc_haveSWRadio";
-                exceptions[] = {};
+                condition = "(([] call TFAR_fnc_haveSWRadio)||([] call TFAR_fnc_haveLRRadio))";
+                exceptions[] = {"isNotInside","isNotSitting"};
                 statement = "";
                 icon = QPATHTOF(ui\ACE_Interaction_Radio_Icon.paa);
                 insertChildren = "[_player] call TFAR_fnc_addRadiosToACE";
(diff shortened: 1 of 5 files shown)


dedmen:
The stereo action only shows the 2 stereo actions that are not currently active right?
So if I am on LEFT then it will only show Center and Right?

Dorbedo:
Exactly.

dedmen:
If we add more actions for the local radio the stereo settings should be in a submenu. So that
Equipment -> Radios -> PRC148->Stereo setting->Left

addons/core/ACEActions.hpp
```diff
@@ -5,23 +5,23 @@ class CAManBase: Man {
             class TFAR_Radio {
                 displayName = "Radios";
                 condition = "(([] call TFAR_fnc_haveSWRadio)||([] call TFAR_fnc_haveLRRadio))";
-                exceptions[] = {"isNotInside","isNotSitting"};
+                exceptions[] = {"isNotInside", "isNotSitting", "isNotSwimming"};
```
dedmen: We want that while swimming tho.. so "isNotSwimming" ? Or am I confusing stuff?

Dorbedo: This way the `isNotSwimming` condition is not evaluated. This way, you can be swimming while interacting.
Ref:
https://github.com/acemod/ACE3/blob/master/addons/medical/ACE_Medical_SelfActions.hpp#L12

Dorbedo:
Can be merged from my side.

Issue #1300, opened 2017-08-13
ai hearing not working for respawned player


sbng-PAPA:
**Arma 3 Version:** `1.74` (stable)
**CBA Version:** `3.4.0` (stable)
**Task Force Radio Version:** `1.-1.0.245`

**Mods:** cba, ace3, tfar

**Description:**
once died or respawned, ai_hearing stop working. just stop.
i'm not technician. sorry :(

dedmen:
Thanks. I know what's wrong. Actually that bug breaks even more than just respawn

dedmen:
Not fixed yet ;) I just know what the issue is.

dedmen:
https://github.com/michail-nikolaev/task-force-arma-3-radio/pull/1326

PR #1326, opened 2017-10-24
Convert AI-Hearing into CBA-Setting

Dorbedo:
* moves AI_Hearing from optional into the main
* adds a CBA-Setting
* disabled by default


addons/ai_hearing/XEH_PreInit.sqf
```diff
@@ -0,0 +1,17 @@
+#include "script_component.hpp"
+
+#include "XEH_PREP.sqf"
+
+[
+    "TFAR_AICanHearPlayer",
+    "CHECKBOX",
+    [LSTRING(SETTING_HEADER), LSTRING(SETTING_DESC)],
+    "Task Force Arrowhead Radio",
```
dedmen: Don't we have a stringtable entry for that?

addons/ai_hearing/functions/fnc_initializeAIHearing.sqf
```diff
@@ -3,21 +3,25 @@
...
 */
 
-if (!hasInterface) exitWith {}; //Only on clients
+if ((!hasInterface)||{!TFAR_AICanHearPlayer}||{missionNamespace getVariable [QGVAR(isInitialized),false]}) exitWith {}; //Only on clients
+
+GVAR(isInitialized) = true;
+
 
 ["TFAR_AI_Detection", "OnSpeak", {
+    If !(TFAR_AICanHearPlayer) exitWith {};
```
dedmen: I would rather just have the setting call removeEventhandler when it get's disabled.

dedmen: But this event only fires on the player so.. Probably doesn't really matter if we have an additional check here.
But Setting should still removeEH if GVAR(isInitialized)

Dorbedo: I would not recomend this to be enabled and disabled during the mission anyways.
I'll change it.

addons/ai_hearing/XEH_PreInit.sqf
Dorbedo: This is the not localized Category Name.

dedmen: Welll.. On the other end it's a Name.
Names usually don't get translated

Dorbedo:
done

dedmen:
I'll leave it open till I'm ready to make the next pre-release

dedmen:
Using the CBA eventhandler also fixes https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/1300
Which was caused because the old one checked against the `player` from before respawn.

PR #1330, opened 2017-11-02
Updating Markdown syntax on all READMEs

jeffreylees:
And small adjustments to copy/arrangement in the English one in specific (because I know English and not the others).

The Markdown adjustments will allow the READMEs to correctly display their content, resulting in documentation that looks and reads much better.


dedmen:
Almost all of these Readmes were already deleted.
We only accept PR's to 1.0 branch. it will overwrite everything on master.

jeffreylees:
Gotcha, will take a look there

PR #1331, opened 2017-11-05
CBA-Settings and 3DEN Attributes

Dorbedo:
- deprecate Modules
- CBA-Settings extended
- new function `TFAR_fnc_parseFrequenciesInput`
the function handles the possible user-inputs for frequencies and clears them from unusable stuff
`["12.1","25"]` can be handled and `12.1,25` or any possible combination
- 3DEN Attributes for default group and unit frequencies
- `TFAR_fnc_generateFrequencies` optimization
- new function `TFAR_fnc_getDefaultRadioSettings`
- adds a possibility to disable the usage of radio-codes


addons/core/functions/fnc_getSwSettings.sqf
```diff
@@ -24,24 +24,33 @@ params[["_radio","",[""]]];
 private _value = TFAR_RadioSettingsNamespace getVariable _radio;
 if (!isNil "_value") exitWith {_value};
 
-
-
-if (!(TF_use_saved_sw_setting) or (isNil "TF_saved_active_sw_settings")) then {
+if ((isNil "TF_saved_active_sw_settings") || {!TF_use_saved_sw_setting} then  {
     private _defaultRadios = call TFAR_fnc_getDefaultRadioClasses;
     private _parent = getText (configFile >> "CfgWeapons" >> _radio >> "tf_parent");
     if ((_defaultRadios select 1) == _parent or {(_defaultRadios select 2) == _parent}) then {
```
dedmen: Could maybe move all this into fnc_getDefaultFrequencies. Which takes a bool as parameter to differentiate between SR and LR.

addons/core/functions/modules/fnc_initialiseFreqModule.sqf
```diff
@@ -25,37 +25,57 @@ params [
...
-
-            if (!isNil (_x getVariable "tf_lr_frequency")) then {private _message = format["TFAR - tf_lr_frequency already set, might be assigning a group (%1) to multiple frequency modules. Or frequency modules to multiple units in the same group.", (group _x)];diag_log _message;hint _message;};
-
-            _x setVariable ["tf_sw_frequency", _swFreq, true];
-            _x setVariable ["tf_lr_frequency", _lrFreq, true];
-
-            true;
-        } count _groups;
+    if (_units isEqualTo []) exitWith {
+        private _msg = "TFAR - No units set for Frequency Module";
```
dedmen: This might happen if you join the mission but no one that was assigned to this module joined yet. I guess we can just remove the warning..

addons/core/functions/modules/fnc_initialiseFreqModule.sqf
```diff
@@ -25,37 +25,57 @@ params [
...
+            private _groups = [];
+            {
+                _groups pushBackUnique (group _x);
+                true;
+            } count _units;
+
+            private _swFreq = _logic getVariable "tf_sw_frequency";
+            private _lrFreq = _logic getVariable "tf_lr_frequency";
+            {
+                If ((!(isNil "_swFreq")) && {!(isNil {_x getVariable "tf_sw_frequency"})}) then {
```
dedmen: the isNil state of _swFreq/_lrFreq never changes inside this loop.
Also SW doesn't exist anymore. It's SR now. though I'd keep the variable name on the group for BWC.

addons/core/functions/modules/fnc_initialiseFreqModule.sqf
```diff
@@ -25,37 +25,57 @@ params [
...
+                    _x setVariable ["tf_sw_frequency", _swFreq];
+                };
+                If ((!(isNil "_lrFreq")) && {!(isNil {_x getVariable "tf_lr_frequency"})}) then {
+                    _x setVariable ["tf_lr_frequency", _lrFreq];
+                };
+            } count _groups;
+        };
+        If (isNil {_logic getVariable "tf_lr_frequency"}) then {
```
dedmen: Are you sure that's possible?

addons/core/functions/fnc_getDefaultRadioSettings.sqf
```diff
@@ -0,0 +1,69 @@
...
+        ARRAY
+
+    Example:
+        _swFrequencies =  call TFAR_fnc_getDefaultRadioSettings;
+        _lrFrequencies = false call TFAR_fnc_getDefaultRadioSettings;
+*/
+
+params [["_isSW", true, [false]], ["_unit", TFAR_currentUnit, [objNull]]];
+
+If (_isSW) then {
```
dedmen: `if` is lowercase.

addons/core/functions/modules/fnc_initialiseFreqModule.sqf
Dorbedo: I was. What's your problem with this?

addons/core/functions/modules/fnc_initialiseFreqModule.sqf
Dorbedo: changed

addons/core/functions/modules/fnc_initialiseFreqModule.sqf
Dorbedo: removed

Dorbedo:
The first test was positive. The module and setting work like expected.

addons/core/XEH_preInit.sqf
```diff
@@ -284,8 +268,190 @@ If (isServer && {isMultiplayer || is3DENMultiplayer}) then {
...
+[
+    "TFAR_setting_DefaultRadio_Rifleman_West", 
+    "EDITBOX", 
+    ["Default Rifleman radio","Default RiflemanRadio west description"], 
+    "Task Force Arrowhead Radio", 
+    "TFAR_rf7800str",
+    1,
+    {
+        TFAR_DefaultRadio_Rifleman_West = If (_this call TFAR_fnc_isPrototypeRadio) then {_this} else {"TFAR_rf7800str"};
+        publicVariable "TFAR_DefaultRadio_Rifleman_West";
```
dedmen: publicVariable here is useless. As that function will be executed on each client and server anyway.

addons/core/XEH_preInit.sqf
```diff
@@ -284,8 +268,190 @@ If (isServer && {isMultiplayer || is3DENMultiplayer}) then {
...
+] call CBA_Settings_fnc_init;
+[
+    "TFAR_setting_DefaultRadio_Rifleman_West", 
+    "EDITBOX", 
+    ["Default Rifleman radio","Default RiflemanRadio west description"], 
+    "Task Force Arrowhead Radio", 
+    "TFAR_rf7800str",
+    1,
+    {
+        TFAR_DefaultRadio_Rifleman_West = If (_this call TFAR_fnc_isPrototypeRadio) then {_this} else {"TFAR_rf7800str"};
```
dedmen: `if` is lowercase.

addons/core/functions/fnc_isBackpackRadio.sqf
```diff
@@ -23,4 +23,6 @@ params [["_classname",[]]];
 if (_classname isEqualType []) exitWith {false};
 if (_classname isEqualType objNull) then {_classname = typeOf _classname;};
 
-([_classname, "tf_hasLRradio", 0] call TFAR_fnc_getConfigProperty) isEqualTo 1
+If (([_classname, "tf_hasLRradio", 0] call TFAR_fnc_getConfigProperty) isEqualTo 0) exitWith {false};
```
dedmen: `if` lowercase

addons/core/XEH_preInit.sqf
Dorbedo: You're right, I've changed it.

addons/core/XEH_preInit.sqf
```diff
@@ -189,9 +189,10 @@ If (isServer && {isMultiplayer || is3DENMultiplayer}) then {
     "",
     1,
     {
-        TFAR_defaultFrequencies_sr_west = [_this, TFAR_MAX_CHANNELS, TFAR_MAX_SW_FREQ, TFAR_MIN_SW_FREQ, TFAR_FREQ_ROUND_POWER] call DFUNC(parseFrequenciesInput);
-        publicVariable "TFAR_defaultFrequencies_sr_west";
-        TFAR_setting_defaultFrequencies_sr_west = str(TFAR_defaultFrequencies_sr_west apply {parseNumber _x});
+        if (isServer) then {
+            TFAR_defaultFrequencies_sr_west = [_this, TFAR_MAX_CHANNELS, TFAR_MAX_SW_FREQ, TFAR_MIN_SW_FREQ, TFAR_FREQ_ROUND_POWER] call DFUNC(parseFrequenciesInput);
+            publicVariable "TFAR_defaultFrequencies_sr_west";
```
dedmen: Currently the biggest problem with default frequencies is that publicVariables don't arrive soon enough on the client.
Instead of doing this on the server and trying to push it to the client and hoping it arrives soon enough.. Just do it on everyone including the client.

Dorbedo: My Problem with this solution is, that the clients could get different settings.
e.g.
The mission defines only the first 2  frequencies.
Then everyone fills the predefined frequencies with his/her own frequencies.
I consider this as not wanted behavior.

I would prefer if the clients don't do anything until the at least the "CBA_settingsInitialized" Event was raised.

dedmen: The code will only be executed when the settings are done.
And afaik it waits till it is sure that the server settings arrived if there are any.

Dorbedo:
Still WIP, do not merge.

Dorbedo:
should be done

addons/core/CBA_Settings.sqf
```diff
@@ -0,0 +1,474 @@
...
+    "SLIDER",
+    ELSTRING(settings,default_radioVolume),
+    "Task Force Arrowhead Radio",
+    [1, 9, 7, 0],
+    0
+] call CBA_Settings_fnc_init;
+[
+    "TFAR_volumeModifier_forceSpeech", 
+    "CHECKBOX", 
+    [ELSTRING(settings,directSpeechModifier), ELSTRING(settings,directSpeechModifier_desc)],
```
dedmen: ELSTRING is correct? You had some typo in the last PR.

addons/core/functions/fnc_getLrRadioCode.sqf
```diff
@@ -22,4 +22,6 @@
     (call TFAR_fnc_activeLrRadio) call TFAR_fnc_getLrRadioCode;
 */
 
+If (GVARMAIN(radioCodesDisabled)) exitWith {""};
```
dedmen: I think this might be called before CBA settings are initialized.
getLrSettings should not initialize till CBA init is done. Don't know what it should do in that case though.. Maybe return empty array...

addons/core/functions/fnc_parseFrequenciesInput.sqf
```diff
@@ -0,0 +1,62 @@
...
+if !(((_valueString find "[") isEqualTo 0) && {(_valueString find "]") isEqualTo ((count _valueString) - 1)}) then {
+    _valueString = format["[%1]",_valueString];
+};
+
+private _parsedValue = (parseSimpleArray _valueString) apply {
+    if (IS_STRING(_x)) then {parseNumber _x} else {_x};
+};
+
+_parsedValue = _parsedValue select {
+    IS_NUMBER(_x) &&
```
dedmen: This will just remove invalid thingies right? We might want to insert a random value. Instead of moving the ones after the invalid one to the front.
If user has 1,2,3 set and 4 invalid and 5,6,7 set.
then their 5 frequency will all of the sudden land on 4.

I don't care though. Just a thing you could do if you feel like it.

addons/core/functions/fnc_radioReplaceProcess.sqf
```diff
@@ -19,32 +19,24 @@
         [TFAR_fnc_radioReplaceProcess, 2] call CBA_fnc_addPerFrameHandler;
 */
 
-if !(TF_use_saved_sw_setting) then {
```
dedmen: This file shouldn't do anything when CBA settings are not there yet.
This is the place to wait at to not initialize SR radios.

addons/core/CBA_Settings.sqf
Dorbedo: `ELSTRING` inside of `*.sqf` `ECSTRING` inside of configs

addons/core/functions/fnc_getLrRadioCode.sqf
Dorbedo: I don't think this has a bigger impact then you could be able to hear something while you are joining.
If I see this right, the result of this function is nowhere stored for a long time.

addons/core/functions/fnc_parseFrequenciesInput.sqf
Dorbedo: In my opinion, the mission-designer should just fix it.

The target of this function was just to prevent invalid stuff.

addons/core/functions/fnc_radioReplaceProcess.sqf
```diff
@@ -19,6 +19,8 @@
         [TFAR_fnc_radioReplaceProcess, 2] call CBA_fnc_addPerFrameHandler;
 */
 
+If !(GVAR(SettingsInitialized)) exitWith {};
```
dedmen: `if` `If` :u

addons/core/functions/fnc_getLrRadioCode.sqf
dedmen: It is stored in the radio settings array.

addons/core/stringtable.xml
```diff
@@ -1277,6 +1317,264 @@
...
+            <Key ID="STR_TFAR_SETTINGS_DefaultRadioFrequencies_SR">
+                <English>Default SR-settings:</English>
+                <German>Standard Kurzwellen-Frequenzen:</German>
+            </Key>
+            <Key ID="STR_TFAR_SETTINGS_DefaultRadioFrequencies_SR_desc">
+                <English>Define the default SR-settings for this entity</English>
+                <German>Das Festlegen der Standard Kurzwellenfrequenzen für diese Entität.</German>
+            </Key>
+            <Key ID="STR_TFAR_SETTINGS_DefaultRadioFrequencies_LR">
+                <English>Default LR-settings:</English>
```
dedmen: This is confusing AF. Should have been LR-frequencies.

PR #1332, opened 2017-11-07
TFAR_fnc_getConfigWeaponProperty

Dorbedo:
- add the function: `TFAR_fnc_getConfigWeaponProperty`
- enables the config-caching for the Server
- renaming the config caching functions and variables


addons/core/functions/fnc_getConfigWeaponProperty.sqf
```diff
@@ -0,0 +1,53 @@
...
+    NUMBER or TEXT or ARRAY - Result
+
+    Example:
+        [_SrRadio, "tf_prototype", 0] call TFAR_fnc_getConfigWeaponProperty;
+*/
+
+params ["_item", "_property", ["_default", ""]];
+
+private _cacheName = (_item + _property);
+private _cachedEntry = TFAR_ConfigWeaponCacheNamespace getVariable _cacheName;
```
dedmen: Namespace could be GVAR. Don't really needs to though. But It's a Internal only thing so it might aswell.

addons/core/functions/fnc_isPrototypeRadio.sqf
```diff
@@ -22,16 +22,12 @@
...
 if (_this == "ItemRadio") exitWith {true};
 
-
-//This is the same caching as in fnc_getConfigProperty. But we don't need that special stuff for CfgVehicles in getConfigProperty
-//So we implement the same caching for this more slim function.
-
-private _cacheName = (_this+"tf_prototype");
-private _cachedEntry = TFAR_ConfigCacheNamespace getVariable _cacheName;
+private _cacheName = (_this+"tf_prototypebool");
+private _cachedEntry = TFAR_ConfigWeaponCacheNamespace getVariable _cacheName;
```
dedmen: Why do you have a copy of the cache function around the cached function here?

addons/core/functions/fnc_isRadio.sqf
```diff
@@ -20,15 +20,12 @@
         _isRadio = "NotARadioClass" call TFAR_fnc_isRadio;
 */
 
-//This is the same caching as in fnc_getConfigProperty. But we don't need that special stuff for CfgVehicles in getConfigProperty
-//So we implement the same caching for this more slim function.
-
-private _cacheName = (_this+"tf_radio");
-private _cachedEntry = TFAR_ConfigCacheNamespace getVariable _cacheName;
+private _cacheName = (_this+"tf_radiobool");
+private _cachedEntry = TFAR_ConfigWeaponCacheNamespace getVariable _cacheName;
```
dedmen: Same here? Why not use the cache func twice?

dedmen:
Not sure about the name
getWeaponConfigProperty or getConfigPropertyWeapon.
We can also rename the old vehicle function so they match in Naming style. That stuff is not public API.

addons/core/functions/fnc_isPrototypeRadio.sqf
Dorbedo: because this way, i can return a bool without checking a number against a bool.

Dorbedo:
done

PR #1336, opened 2017-11-10
Stringtable - splitting and macros

Dorbedo:
- splits the stringtable into one for each component
- usage of cba string macros
- some new strings
- some old strings removed
- some strings marked as depracted to be removed in another PR


Dorbedo:
done

dedmen:
The best thing since sliced bread

PR #1338, opened 2017-11-12
Fixes for 268

Dorbedo:
* old functionname was used
* settings string was misdone
* `TFAR_fnc_sendPluginConfig` runs only if the settings are already initialized
* Keybinds were not localized
* wrong stringname
* supports Channelname and password set before postinit


Dorbedo:
My merging fault.

addons/core/XEH_preInit.sqf
```diff
@@ -16,8 +16,8 @@ GVAR(WeaponConfigCacheNamespace) = false call CBA_fnc_createNamespace;
 if (hasInterface) then {//Clientside Variables
     call TFAR_fnc_initKeybinds;
     //PreInit variablesy
-    VARIABLE_DEFAULT(tf_radio_channel_name,"TaskForceRadio");
-    VARIABLE_DEFAULT(tf_radio_channel_password,"123");
+    VARIABLE_DEFAULT(TFAR_radio_channel_name,"TaskForceRadio");
```
dedmen: If you wanted to go crazy... Check inside the onSettingChanged function of the CBA setting if TFAR_radio_channel_name already exists and don't overwrite it if it does.
That way users can manually set it in scripts before or after CBA settings are initialized.

Dorbedo: But you couldn't update your own settings then.

Dorbedo:
```
// init.sqf
[
    "CBA_settingsInitialized",
    {
        TFAR_Teamspeak_Channel_Name = "My_Channel";
        TFAR_Teamspeak_Channel_Password = "My_Password";
        [] call TFAR_fnc_sendPluginConfig;
        ["CBA_settingsInitialized", _thisId] call CBA_fnc_removeEventHandler;
    }
] call CBA_fnc_addEventHandlerArgs;
```
This should be enough, if you want to set them by your own.

dedmen: Problem is that init.sqf might run after CBA_settingsInitialized

Dorbedo: Shouldn't since it's called postinit:
https://github.com/CBATeam/CBA_A3/blob/master/addons/settings/XEH_postInit.sqf

dedmen: init.sqf can run minutes after postInit.. So.. No.

Dorbedo: Alright then. Just add it to preInit.

After this PR, the users will still be able to use the old variable-names to overwrite the Settings.
Otherwise they could wait until `tfar_core_settingsInitialized == true` before they overwrite the setting provided by the 3DEN Attribute and refresh the plugin config afterwards.

dedmen:
Gimme a go and I'll give it a go

Dorbedo:
go

PR #1339, opened 2017-11-13
Update the Jpn translation

classicarma:
- support to new entry
- removed double entry from core stringtable
- update, fix and tweak for translation


addons/core/stringtable.xml
```diff
@@ -1,26 +1,31 @@
-﻿<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
dedmen: I'd prefer if that line would not be changed.

addons/core/stringtable.xml
```diff
@@ -1171,139 +1193,164 @@
...
                 <Chinesesimp>当无线电频道更改时在右下角显示提示讯息</Chinesesimp>
+                <Japanese>UI か割り当てキーでチャンネルを変更するとヒント (右下) を表示します</Japanese>
             </Key>
             <Key ID="STR_TFAR_SETTINGS_FullDuplex">
                 <English>Full-Duplex</English>
                 <French>Duplex complet</French>
                 <German>Voll-Duplex</German>
                 <Japanese>全二重通信</Japanese>
                 <Korean>양방향 통신</Korean>
-                <English>Full-Duplex</English>
```
dedmen: Shouldn't stringtables have a <original> entry?

addons/core/stringtable.xml
```diff
@@ -1817,53 +1906,53 @@ Normal radyo su altı radyosuyla iletişim kuramaz (herhangi bir ses gelmez). Da
             </Key>
             <Key ID="STR_TFAR_CORE_radio_control_text">
                 <English>&lt;br/&gt;
-&lt;font color="#67ED24"&gt;Push-to-talk button in TeamSpeak&lt;/font&gt;	Direct Speech.&lt;br/&gt;
```
dedmen: You don't need to escape quotes.

addons/core/stringtable.xml
Dorbedo: this removal is fine.
If comes from converting original to english

dedmen: I know it's fine. But what does Arma do if there is no <original> entry? Does it automatically use english or the first entry or at random?

addons/handhelds/stringtable.xml
```diff
@@ -3,61 +3,75 @@
     <Package name="backpacks">
         <Key ID="STR_TFAR_handhelds_RF7800STR">
             <Original>RF-7800S-TR</Original>
```
Dorbedo: Could you please do me a favor and replace `<Original> *** </Original>` with `<English> *** </English>`
If forgot to change this inside this stringtable

addons/core/stringtable.xml
Dorbedo: the first entry

addons/handhelds/stringtable.xml
```diff
@@ -2,53 +2,53 @@
 <Project ID="TFAR">
     <Package name="backpacks">
         <Key ID="STR_TFAR_handhelds_RF7800STR">
-            <Original>RF-7800S-TR</Original>
```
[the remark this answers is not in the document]
dedmen: Also I prefer the original translation to be "Original".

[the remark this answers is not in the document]
Dorbedo: Keep it this way, I'll change it in another PR.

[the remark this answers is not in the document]
classicarma: So, what should I do? Keep it?

PR #1341, opened 2017-11-18
Static Command Station radio

JTLeClair:
I had suggested this feature on the BiStudio forums and was asked to put the request on GitHub. Our HQ element usually runs command and control from a Military Cargo HQ, and it would be interesting to have a static radio/command station for the radio operator of the CoC element to sit at while inside the trailer instead of having to have a backpack radio on at all times. I don't know if a static model that works similar to a vehicle ("enter" it to sit down, radio works like a vehicle radio) would be the best option, or if this would be something you would develop alongside your work getting backpack radios usable by other people around the carrier. 

Also, I still don't know how to navigate GitHub that well, so I may have posted this in the wrong place.


dedmen:
This is a pull request.
Issues and feature requests go to https://github.com/michail-nikolaev/task-force-arma-3-radio/issues

PR #1345, opened 2017-11-22
TS-Channelname changed into CBA_settings

Dorbedo:
- remove the 3DEN Channelsetting
- add the CBA-Setting for Channelname and PW
- needs next CBA

### DIFF
--- a/addons/core/CBA_Settings.sqf
+++ b/addons/core/CBA_Settings.sqf
@@ -160,6 +160,22 @@
     {["spectatorCanHearFriendlies", TFAR_spectatorCanHearFriendlies] call TFAR_fnc_setPluginSetting;}
 ] call CBA_Settings_fnc_init;
 [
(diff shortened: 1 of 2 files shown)


dedmen:
waiting for password masking thingy in CBA.

dedmen:
Can you add https://github.com/CBATeam/CBA_A3/pull/802/files ?
Then this can be merged-

Dorbedo:
It's done, just need cba 3.5.1 to be released

dedmen:
Does it break if we merge it before CBA release?

Dorbedo:
Yes, it would break.

dedmen:
https://github.com/CBATeam/CBA_A3/blob/release/addons/settings/fnc_init.sqf#L91
That looks like they just use the first array element though.. Which would be `""` which would work just fine?

PR #1351, opened 2017-12-05
cba_settings file renamed, fixes

Dorbedo:
- renaming cba_settings.sqf -> fnc_initCBASettings.sqf
- fix a typo
- FullDuplex gets now set if changed during mission


Dorbedo:
done

PR #1352, opened 2017-12-05
settings categories

Dorbedo:
- Divides the settings into 2 categories
- Forces some settings to be only set by clients


Dorbedo:
this is done.
If the next CBA is released, we could drop the `localize`

PR #1357, opened 2017-12-14
bwc for tf_guer_radio_code

Dorbedo:
`tf_guer_radio_code` should stay for this version instead of `tf_independent_radio_code`

### DIFF
--- a/addons/backpacks/anarc164/CfgVehicles.hpp
+++ b/addons/backpacks/anarc164/CfgVehicles.hpp
@@ -10,7 +10,7 @@ class TFAR_anarc164: TFAR_Bag_Base {
     maximumLoad = 20;
     mass = 160;
     tf_range = 40000;
-    tf_encryptionCode = "tf_independent_radio_code";
+    tf_encryptionCode = "tf_guer_radio_code";
     tf_dialog = "anarc164_radio_dialog";
     tf_subtype = "airborne";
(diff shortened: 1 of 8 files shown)


dedmen:
You are missing one.

dedmen:
https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/97583dc6a8a42c958460d3b2d7eca57469fe7757/arma3/%40task_force_radio/addons/task_force_radio/functions/fn_ClientInit.sqf#L19 That was maybe the reason why it was changed. Not sure

https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/47d03aa686d96169fe13a5c4cfe139b081031265/arma3/%40task_force_radio/addons/task_force_radio/functions/fn_initialiseBaseModule.sqf#L66 There. Is that code still somewhere?

dedmen:
guer was used because `independent` stringified as GUER in A2. But I think that changed in A3 and I think that stringification caused some problems. So I probably changed it so it matches

Dorbedo:
I'll take a deeper look another time, so no merge for now.

dedmen:
Keep in mind that I got very little sleep.

Dorbedo:
alright, it's reverted and changed into supporting of the old value

dedmen:
Get sleep o7

Dorbedo:
rdy to merge

PR #1360, opened 2017-12-15
ai-hearing improvements

Dorbedo:
- change the remoteExec into a CBA-Event
- some of the TODO
- cleanup
- reduced the remote calling by adding some additional checks
- new setting to enable the ai abillity to detect speakers on receive (has some limitations)

### DIFF
--- a/addons/ai_hearing/XEH_PREP.sqf
+++ b/addons/ai_hearing/XEH_PREP.sqf
@@ -1,4 +1,4 @@
-PREP(OnRadioReceive);
+PREP(onRadioReceive);
 PREP(onSettingChanged);
 PREP(onSpeak);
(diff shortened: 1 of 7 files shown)


dedmen:
That is actually a thing remoteExec is useful for. Directly calling the command is ofcause better than CBA parsing the event PVAR and calling a script function that calls a command

Dorbedo:
I know, but this will work even if there are strange things inside `CfgRemoteExec`.

Dorbedo:
should be done

dedmen:
rdy rdy rdy?

Dorbedo:
can be merged

dedmen:
splendid

PR #1361, opened 2017-12-15
acenal

Dorbedo:
- supports the saving of baseclass inside saved loadouts (only ACEnal)


dedmen:
Completly missed that they implemented it

PR #1367, opened 2017-12-18
TFAR_fnc_processPlayerPositions - cleanup&opti

Dorbedo:
- cleanup TFAR_fnc_processPlayerPositions 
- do some small optimization inside TFAR_fnc_processPlayerPositions

### DIFF
--- a/addons/core/functions/fnc_processPlayerPositions.sqf
+++ b/addons/core/functions/fnc_processPlayerPositions.sqf
@@ -18,40 +18,38 @@
     Example:
         call TFAR_fnc_processPlayerPositions;
 */
-if (getClientStateNumber != 10) exitWith {"BI HAS CRAPPY WEIRD BUGS U KNOW! (Keeps PFH from firing after server disconnect)"};
-//if (isNull (findDisplay 46)) exitWith {};
-//if (isNull TFAR_currentUnit) exitWith {};
+if (getClientStateNumber != 10) exitWith {};
+
 private _startTime = diag_tickTime;
 
 //Process queued Near Players
 if (!TFAR_currentNearPlayersProcessed) then {
+
+    if (TFAR_currentNearPlayersProcessing isEqualTo []) exitWith {TFAR_currentNearPlayersProcessed = true};
+
     private _nearPlayersCount = count TFAR_currentNearPlayersProcessing;
-    if (_nearPlayersCount == 0) exitWith {TFAR_currentNearPlayersProcessed = true};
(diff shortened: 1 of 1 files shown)


addons/core/functions/fnc_processPlayerPositions.sqf
```diff
@@ -18,80 +18,84 @@
     Example:
         call TFAR_fnc_processPlayerPositions;
 */
-if (getClientStateNumber != 10) exitWith {"BI HAS CRAPPY WEIRD BUGS U KNOW! (Keeps PFH from firing after server disconnect)"};
-//if (isNull (findDisplay 46)) exitWith {};
-//if (isNull TFAR_currentUnit) exitWith {};
+if (getClientStateNumber != 10) exitWith {};
```
dedmen: Awwwww I liked that

addons/core/functions/fnc_processPlayerPositions.sqf
```diff
@@ -18,80 +18,84 @@
...
             [_controlled, true, _unitName] call PROFCONTEXT_NORTN(TFAR_fnc_sendPlayerInfo);
         };
-    } forEach (TFAR_currentNearPlayersProcessing select [0,_playersToProcess]); //commy2
-
-    //private _startIndex = _nearPlayersCount - _playersToProcess; //Is min 0
-    //for "_y" from _startIndex to _nearPlayersCount-1 step 1 do {
-    //    private _unit = (TFAR_currentNearPlayersProcessing select _y);
-    //};
+        nil
+    } count (TFAR_currentNearPlayersProcessing select [0, _playersToProcess]); //commy2
```
dedmen: count+extra command can be slower than a simple forEach.
nil isn't needed here because it would use the returned value from sendPlayerInfo

Dorbedo: `count` is always faster then `foreach`.
no `_forEachIndex` every iteration -> it's faster

dedmen: Yes. But if you add an additional commands thats 3 microseconds slower on each iteration. versus the about 1 microsecond that count is faster.

addons/core/functions/fnc_processPlayerPositions.sqf
```diff
@@ -18,7 +18,7 @@
     Example:
         call TFAR_fnc_processPlayerPositions;
 */
-if (getClientStateNumber != 10) exitWith {};
+if !(getClientStateNumber isEqualTo 10) exitWith {};
```
dedmen: No not faster. 2 script command calls now instead of just 1 before.

Dorbedo:
can be merged

PR #1371, opened 2017-12-31
Add getVehicleRadios function to allow easy setup of radios on vehicles.

DeliciousJaffa:


addons/core/functions/fnc_getVehicleRadios.sqf
```diff
@@ -0,0 +1,37 @@
...
+        Jaffa
+
+    Description:
+        Gets a list of radios in the specified vehicle
+
+    Parameters:
+        0: Object - vehicle
+
+    Returns:
+         Array of radios in format:
```
dedmen: Array of Arrays?

addons/core/functions/fnc_getVehicleRadios.sqf
```diff
@@ -0,0 +1,37 @@
...
+         Array of LR radio arrays in format: 
+             0 - Object - Vehicle, 1 - String - Radio Settings ID
+         or empty array if no vehicle radios found
+
+    Example:
+        _radios = _vehicle call TFAR_fnc_getVehicleRadios;
+*/
+params ["_vehicle"];
+
+if (isNull _vehicle || {!(_vehicle call TFAR_fnc_hasVehicleRadio)}) exitWith {[]};//Vehicle is null or vehicle doesn't have LR Radio
```
Dorbedo: The comment in this line is useless

addons/core/functions/fnc_getVehicleRadios.sqf
```diff
@@ -0,0 +1,37 @@
...
+
+private _result = [[_vehicle, "gunner_radio_settings"],[_vehicle, "driver_radio_settings"],[_vehicle, "commander_radio_settings"],[_vehicle, "copilot_radio_setting"]];
+        
+private _turrets = [(typeof _vehicle), "TFAR_AdditionalLR_Turret", []] call TFAR_fnc_getVehicleConfigProperty;
+private _turrets = _turrets apply {[_vehicle, format ["turretUnit_%1_radio_setting",_x]]};
```
Dorbedo: Don't private again.

addons/core/functions/fnc_getVehicleRadios.sqf
```diff
@@ -0,0 +1,37 @@
...
+_result append _turrets;
+
+private _cargos = [(typeof _vehicle), "TFAR_AdditionalLR_Cargo", []] call TFAR_fnc_getVehicleConfigProperty;
+private _cargos = _cargos apply {[_vehicle, format ["cargoUnit_%1_radio_setting",_x]]};
```
Dorbedo: Don't private again.

addons/core/functions/fnc_getVehicleRadios.sqf
Dorbedo: Ohh, and this is wrong:
`TFAR_AdditionalLR_Turret` returns the following (example):
https://git.koffeinflummi.de/bwmod-team/bwa3-public/blob/master/bwa3_leopard2a6m/CfgVehicles.hpp#L71
This would result in:
`"turretUnit_[0,1]_radio_setting"` instead of e.g.: "turretUnit_0_radio_setting"
ref: https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/1.0/addons/core/functions/fnc_vehicleLr.sqf#L45

addons/core/functions/fnc_getVehicleRadios.sqf
Dorbedo: same format issue as above

dedmen: Not same format issue: https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/1.0/addons/core/functions/fnc_vehicleLr.sqf#L49

Dorbedo: yes, my fault. I didn't check enough.
The format is right here, the private not.

DeliciousJaffa:
Both issues should be fixed now, didn't realise the turrets were index based.

addons/core/functions/fnc_getVehicleRadios.sqf
DeliciousJaffa: Tell Nkey that https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/1.0/addons/core/functions/fnc_vehicleLr.sqf#L25

Dorbedo: I know, that the comment is in this function too.
But it's still useless, because it's not providing any additional information.
It's just the same as the code.

addons/core/functions/fnc_getVehicleRadios.sqf
```diff
@@ -0,0 +1,36 @@
...
+
+    Returns:
+         Array of LR radio arrays in format: 
+             0 - Object - Vehicle, 1 - String - Radio Settings ID
+         or empty array if no vehicle radios found
+
+    Example:
+        _radios = _vehicle call TFAR_fnc_getVehicleRadios;
+*/
+params ["_vehicle"];
```
Dorbedo: I think, you could add a default value and a required datatype since this function is considered as API/Public stuff:
`params [["_vehicle", objNull, [objNull]]];`

DeliciousJaffa:
@Dorbedo Latest commit will type check and give a useful error if an incorrect type or no argument is supplied.

addons/core/functions/fnc_getVehicleRadios.sqf
```diff
@@ -0,0 +1,36 @@
...
+    Returns:
+         Array of LR radio arrays in format: 
+             0 - Object - Vehicle, 1 - String - Radio Settings ID
+         or empty array if no vehicle radios found
+
+    Example:
+        _radios = _vehicle call TFAR_fnc_getVehicleRadios;
+*/
+params [["_vehicle",0]];
+if !(_vehicle isEqualType objNull) exitWith { ["TFAR: Vehicle must be passed to getVehicleRadios. %1 was passed instead.",_vehicle] call BIS_fnc_error };
```
dedmen: Use CBA error thingy

addons/core/functions/fnc_getVehicleRadios.sqf
```diff
@@ -0,0 +1,36 @@
...
+
+    Returns:
+         Array of LR radio arrays in format: 
+             0 - Object - Vehicle, 1 - String - Radio Settings ID
+         or empty array if no vehicle radios found
+
+    Example:
+        _radios = _vehicle call TFAR_fnc_getVehicleRadios;
+*/
+params [["_vehicle",0]];
```
dedmen: Doesn't typeCheck that you pass an object

DeliciousJaffa: It does here https://github.com/michail-nikolaev/task-force-arma-3-radio/pull/1371/files#diff-07960b095b64fe84277c6f2cf1e09662R24

Dorbedo: Why not use the return value of params:
`If !(params [["_vehicle", objNull, [objNull]]]) exitWith {/*ERROR*/};`

DeliciousJaffa:
@dedmen Changed the error to use CBA's ERROR macro

addons/core/functions/fnc_getVehicleRadios.sqf
```diff
@@ -0,0 +1,36 @@
...
+if !(_vehicle isEqualType objNull) exitWith { ERROR(format ["Vehicle must be passed to getVehicleRadios. %1 was passed instead.",_vehicle]) };
```
Dorbedo: `ERROR(format ["Vehicle must be passed to getVehicleRadios. %1 was passed instead.",_vehicle])`
won't compile right.
Use:
`ERROR_1("Vehicle must be passed to getVehicleRadios. %1 was passed instead.",_vehicle)`

addons/core/functions/fnc_getVehicleRadios.sqf
```diff
@@ -0,0 +1,35 @@
...
+
+    Returns:
+         Array of LR radio arrays in format: 
+             0 - Object - Vehicle, 1 - String - Radio Settings ID
+         or empty array if no vehicle radios found
+
+    Example:
+        _radios = _vehicle call TFAR_fnc_getVehicleRadios;
+*/
+if !(params [["_vehicle",objNull,[objNull]]]) exitWith { ERROR_1("TFAR: Vehicle must be passed to getVehicleRadios. %1 was passed instead.", _vehicle) };
```
Dorbedo: You could return an empty array even if the wrong input was given.

DeliciousJaffa: There's no use case in which a non-object would need to be passed into the function.

dedmen:
@Dorbedo want this merged before your header updates? so you can include that new functions header too? Or just directly get it fixed in here?

Dorbedo:
no, merge the header first, I'll create another small PR

PR #1382, opened 2018-02-01
Added Material icons

chris579:
Needs to be discussed. All icons are licensed under Apache 2.0.


dedmen:
>licensed under Apache 2.0

https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/master/LICENSE.md

Sooo.....

654wak654:
I do like the new icons but an extra clause to the license should be added stating the icons are under Apache 2.0 and not APL-SA.

Something like this maybe? https://github.com/acemod/ACE3/blob/master/LICENSE#L25-L31

Dorbedo:
The icons are not at the same position -> they are all centered instead
This results in an ugly "jumping" of the icon upon switch

dedmen:
We also don't need the png's if we have the .svg. The png's are just there to have the source files somewhere. I prefer svg.

nomisum:
I could provide juggle free alignment of those. I presume SVG is not the format used by Arma though?

dedmen:
Yes. Arma also doesn't use PNG.

nomisum:
how about those
tfar_volume

dedmen:
These are actually perfect XD

nomisum:
License is "do what you want", as they are mine :D Will provide you the sources tomorrow.

dedmen:
Can we get headset into that too?

PR #1395, opened 2018-04-05
Add New Icons

654wak654:
Add new icons made by @nomisum.


dedmen:
Don't need nor want the png's. Everything here will be packed into PBO. And there is just no need for the png's in there.

dedmen:
The .paa's are not Git LFS :U

654wak654:
I kept getting errors so I just brute forced it. Couldn't you move them to LFS after merging?

dedmen:
No. That wouldn't make sense. Because at that points the files are already in the history as non-LFS. Which is exactly what i don't want ^^
I can also do it in your PR for you. Remind me monday ^^

654wak654:
@dedmen Screw breakfast, start your week by fixing other people's PRs! :)

dedmen:
I don't eat breakfast usually.

dedmen:
I can't even checkout your branch because Gitkraken fails on it because git complains about
```
Encountered 6 file(s) that should have been pointers, but weren't:
        addons/core/ui/tfar_volume_normal.paa
        addons/core/ui/tfar_volume_normal_hl.paa
        addons/core/ui/tfar_volume_whisper.paa
        addons/core/ui/tfar_volume_whisper_hl.paa
        addons/core/ui/tfar_volume_yelling.paa
        addons/core/ui/tfar_volume_yelling_hl.paa
```
:D

PR #1406, opened 2018-04-26
optimizing getRadioItems

Dorbedo:
fix function header
optimize the function


dedmen:
What do you think. Should we replace all of these by https://github.com/CBATeam/CBA_A3/pull/902?
The seperate function call will make everything slower though..

Dorbedo:
sure, why not.
I'd merge this anyways.

dedmen:
won't use the CBA func because that doesn't have the option to leave duplicates.

PR #1442, opened 2018-10-05
Fix Unsung is overriding the ItemRadio class

shukari:
and set `scopeArsenal = 2` ... so any _* radio is in the ACE Arsenal.


dedmen:
`scope = 1` items should not show up in Arsenal...
But ACE Arsenal ignores the scope entry if scopeArsenal entry exists :U I guess BI Arsenal does the same 🤦‍♂️

PR #1444, opened 2018-10-23
new isTurnedOut function

shukari:
- ~~wait for CBA release 3.9.1~~
- old function never gets used
- new function will make many vehicle positions turnout (https://github.com/CBATeam/CBA_A3/pull/1015)


dedmen:
Need to keep the old function name as deprecated and redirect to the CBA func.

shukari:
Since is stopped working on the PR (https://github.com/CBATeam/CBA_A3/pull/1015) for CBA you can use that function here... i can port it to TFAR if you want.

dedmen:
Yeah I will try porting your CBA function. Might need to modify some parts of that. And i didn't check it's performance yet which is also very important.

dedmen:
Oops. Sorry I keep forgetting that PR's exist.

PR #1445, opened 2018-11-25
add TFAR_OverrideActiveLRRadio support

nomisum:
add tfar_overrideactivelrradio


dedmen:
Nah I wanted to have Override not appear in there. as activeLr is already set to override.

PR #1452, opened 2019-01-15
Fixed transmission not ending on kick/crash/timeout

vabene1111:
When a player is kicked, the server stopps or a player times out while holding down the radio transmit button teamspeak will think that the player ist still transmitting.

As a result of that Teamspeak will continue to show the talking indicator light until teamspeak ist restarted or the player connects again.

This simple fix calls the `releaseAllTangents` command just after the normal `MISSIONEND` plugin command is send to reset the transmission status in teamspeak.
I think this is the best way of solving this problem but there might be better ones. 

I have not found side effekts that could cause any problems with this soulution, but i am no expert with TFAR so there might be problems with this soulution that i have not considerd. 

If this is not wanted for any reason feel free to reject this PR

### DIFF
--- a/addons/core/functions/events/fnc_onMissionEnd.sqf
+++ b/addons/core/functions/events/fnc_onMissionEnd.sqf
@@ -18,6 +18,6 @@
   Public: Yes
 */
 
-"task_force_radio_pipe" callExtension "MISSIONEND~"
+"task_force_radio_pipe" callExtension "MISSIONEND~";
 
 player call TFAR_fnc_releaseAllTangents;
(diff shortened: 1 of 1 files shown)


dedmen:
MISSIONEND should release all tangents inside the plugin.
Mission end means it's over and there's nothing coming anymore

vabene1111:
i thought so too but we have tested it with the latest plugin (one person holding down radio button, the other stopping the server) and the talk indicator stays turned on.

I cant think of what but maybe somethign interfered with our tests, so consider this closed

dedmen:
>MISSIONEND forces instant plugin timeout, moving the player out of the TFAR channel and killing all transmissions.

That is also the reason why your "fix" is bad. We are forcing plugin timeout. And after that your "fix" would immediately re-active the plugin again by sending a message and Teamspeak would think the game is still there and might move you back into TFAR channel immediately.

vabene1111:
ok, that dafinitly makes sense, just could not solve it differently since i havent learned how the plugin works yet :D

Issue #1468, opened 2019-06-04
Ability for "Same frequencies per group"


dedmen:
Each group get's their own set of random frequencies.
BUT the group members all share the same frequency set.
Same as "SameFrequenciesPerSide" just on Group level.

PR #1469, opened 2019-06-19
Same SR frequencies per group

severgun:
- Implement https://github.com/michail-nikolaev/task-force-arma-3-radio/issues/1468

There is quite big chance of collisions so may be it is necessary to add uniqueness test for first element in array, or `TFAR_FREQ_ROUND_POWER` should be set to 100.


addons/core/functions/fnc_initCBASettings.sqf
```diff
@@ -575,18 +583,18 @@
...
+    [ELSTRING(settings,instantiateAtBriefing), ELSTRING(settings,instantiateAtBriefing_desc)],
+    localize ELSTRING(settings,global),
     false,
     1
 ] call CBA_Settings_fnc_init;
 [
-    "TFAR_defaultIntercomSlot", 
-    "SLIDER", 
-    ["Default Intercom Channel", "Default Intercom channel when entering the vehicle. Has to be number. -1 is Disabled."], 
-    localize ELSTRING(settings,global),
```
dedmen: what are these end of line changes?

addons/core/functions/server/fnc_serverInit.sqf
```diff
@@ -113,6 +113,17 @@ GVAR(instanciationIsReady) = false;
...
+
+            if (TFAR_SameSRFrequenciesPerGroup) then {
+                {
+                    _x params ["_unit"];
+                    private _frequencies = _unit getVariable ["TFAR_freq_sr", (group _unit) getVariable "TFAR_freq_sr"];
+                    if (isNil "_frequencies") then {
+                        _frequencies = [TFAR_MAX_CHANNELS, TFAR_MAX_SW_FREQ, TFAR_MIN_SW_FREQ, TFAR_FREQ_ROUND_POWER] call TFAR_fnc_generateFrequencies;
+                        (group _unit) setVariable ["TFAR_freq_sr", _frequencies, true];
+                    };
+                } forEach (playableUnits);
```
dedmen:
```suggestion
                } forEach allPlayers;
```

dedmen: Maybe this should be independent of "TFAR_SameSRFrequenciesForSide".
Otherwise it taking no effect if SameSideFreq's are also enabled is badly documented atleast.

dedmen:
> or instead of TFAR_FREQ_ROUND_POWER value should be set to 100

Would be a compatibility breaker.

I don't care much about uniqeness, we already don't care for the "same frequencies per side" feature.

addons/core/functions/server/fnc_serverInit.sqf
dedmen: Oops, actually this won't work. serverInit runs once. Players might join JIP and their groups then won't get preset frequencies as they only get created when the player joins.

dedmen:
Fixes #1468

addons/core/functions/fnc_initCBASettings.sqf
severgun: Atom done this changes automatically

addons/core/functions/server/fnc_serverInit.sqf
severgun: > Maybe this should be independent of "TFAR_SameSRFrequenciesForSide".

Only one of them can be used at a time. Yes it can confuse.
As you can see in first commit I did it independent of "TFAR_SameSRFrequenciesForSide".

> Oops, actually this won't work.

😞

addons/core/functions/fnc_initCBASettings.sqf
dedmen: https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/master/.editorconfig#L4
:thinking: 
Need to double check that later

addons/core/functions/server/fnc_serverInit.sqf
dedmen: >Oops, actually this won't work.

We need some kind of player joined/group created eventhandler. Definitely possible.

addons/core/functions/fnc_initCBASettings.sqf
severgun: Seems like my fault. 
```
git config --global core.autocrlf
true
```

dedmen:
I like this and this also works fine.
But this only runs on serverInit, which runs once.
If any player JIPs and while doing that creates a new group, this won't execute on his group.
We'd need some kind of playerJoined/groupCreated eventhandler, to run this then.

severgun:
Added event. Will it work now?

addons/core/functions/server/fnc_serverInit.sqf
```diff
@@ -113,6 +127,22 @@ GVAR(instanciationIsReady) = false;
             VARIABLE_DEFAULT(TFAR_freq_sr_west,true call DFUNC(generateSRSettings));
             VARIABLE_DEFAULT(TFAR_freq_sr_east,true call DFUNC(generateSRSettings));
             VARIABLE_DEFAULT(TFAR_freq_sr_independent,true call DFUNC(generateSRSettings));
+
+            if (TFAR_SameSRFrequenciesPerGroup) then {
+                [QGVAR(groupCreated), {
+                    params ["_group"];
+                    _frequencies = [TFAR_MAX_CHANNELS, TFAR_MAX_SW_FREQ, TFAR_MIN_SW_FREQ, TFAR_FREQ_ROUND_POWER] call TFAR_fnc_generateFrequencies;
```
dedmen:
```suggestion
                    private _frequencies = [TFAR_MAX_CHANNELS, TFAR_MAX_SW_FREQ, TFAR_MIN_SW_FREQ, TFAR_FREQ_ROUND_POWER] call TFAR_fnc_generateFrequencies;
```

addons/core/functions/server/fnc_serverInit.sqf
```diff
@@ -113,6 +127,22 @@ GVAR(instanciationIsReady) = false;
...
+                    _group setVariable ["TFAR_freq_sr", _frequencies, true];
+                }] call CBA_fnc_addEventhandler;
+                
+                {
+                    _x params ["_unit"];
```
dedmen: _x is not an array. params doesn't really make sense if you are just doing `_unit = _x`

dedmen:
I don't like the 0 delay PFH, but if there is no better alternative.. I guess it has to be.

addons/core/functions/server/fnc_serverInit.sqf
severgun: seems like copy-paste leftovers. fixed

severgun:
There is no native BIS EH for group creation. So I think that is only one option.
Delay can be increased but then there can be collisions when some groups deleted and some created during delay.

addons/core/functions/server/fnc_serverInit.sqf
dedmen: 2.10 will add group eventhandlers, including join.

dedmen:
> There is no native BIS EH for group creation. 

https://community.bistudio.com/wiki/Arma_3:_Mission_Event_Handlers#GroupCreated
https://community.bistudio.com/wiki/Arma_3:_Mission_Event_Handlers#GroupDeleted

Do you want to update the PR with it or shall I merge this now and upgrade later?

severgun:
I will take a look at this week.

severgun:
Couple questions:
1) "GroupCreated" fires for 33 groups even if only 1 exist.
2) Old feature "group freqs" from group attributes does not work. Seems like attribute expression called before group is fully populated and values are not broadcasted.
3) If "same per group" and "same for side" are selected, what we should take in favor?

dedmen:
1. I think there is something in vanilla code that creates bogus groups. But that was fixed recently-ish in a game update.
2. Mh maybe we can just execNextFrame
3. Closest wins, so group

PR #1471, opened 2019-06-19
Fix Cycle SR/LR keybinds

severgun:
- title


dedmen:
What do these keys even do? :D

severgun:
When you have multiple radios in inventory you can switch them back and forth.
`CTRL+[` , `CTRL+]` for SR by default

dedmen:
Ah. So back and forth (left and right was confusing me), I guess currently only forth worked?

severgun:
Correct. Left and Right confuse me too, but I don't wan't to touch variables naming as much as possible.

dedmen:
Changing the stringtable entries is no issue as they are not referenced anywhere else. And the functions already get called with "next" and "prev" arguments.

severgun:
Like this?
Now it will conflict with https://github.com/michail-nikolaev/task-force-arma-3-radio/pull/1472

PR #1472, opened 2019-06-19
Translation russian

severgun:
- Add missing Russian translation
- Add some missing localization entries in English and Russian (still left some)
- Remove extra colons
- SW to SR in English strings


addons/core/stringtable.xml
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="utf-8"?>
```
dedmen:
```suggestion
<?xml version="1.0" encoding="UTF-8"?>
```
What editor did you use?

severgun: https://github.com/bux/tabler
and Sublime

PR #1478, opened 2019-07-29
Add support for radios in Contact DLC

Dahlgren:
- Add backpack radio support for Contact DLC backpacks

### DIFF
--- a/addons/backpacks/config.cpp
+++ b/addons/backpacks/config.cpp
@@ -32,11 +32,6 @@ class CfgPatches {
     weapons[] = {};
     requiredVersion = REQUIRED_VERSION;
     requiredAddons[] = {
-        "A3_Modules_F",
-        "A3_UI_F",
-        "A3_Structures_F_Items_Electronics",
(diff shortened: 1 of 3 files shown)


addons/backpacks/config.cpp
```diff
@@ -35,6 +35,7 @@ class CfgPatches {
         "A3_Modules_F",
         "A3_UI_F",
         "A3_Structures_F_Items_Electronics",
+        "A3_Supplies_F_Enoch_Bags",
```
dedmen: This whole stuff really isn't needed. tfar_core requires cba_common which requires all of vanilla Arma.
But no need to really fix it either.

dedmen:
Range is missing, I guess we have a default fallback for that?
The config defaults in scripts aren't really trustworthy currently. But I wanted to look into that anyway and add some new defaults. Like that dialogUpdate thing which is always the same.

Dahlgren:
Added `tf_range = 20000;` to all Enoch radios, same as is set for `Bag_Base`

PR #1495, opened 2020-01-06
fnc_haveDDRadio rewritten to use config attributes instead of inheritance

vabene1111:
Changed the function haveDDRadio to use a config attribute instead of inheritance from `V_RebreatherB` to determine whether or not a player has a DD radio.

I tried to follow all naming and style rules i could see, if something should be changed feel free to point it out or change it yourself. Feel free to reject if this is not wanted this way.

- not change any noticable feature to a player
- allow developers to define vests as DD radios without having to inherit `V_RebreatherB`
- probably slightly increase performance of `fnc_haveDDRadio.sqf`

### DIFF
--- a/addons/core/config.cpp
+++ b/addons/core/config.cpp
@@ -17,6 +17,7 @@ class CfgPatches {
             "cba_ui",
             "cba_xeh",
             "cba_settings",
+            "A3_Weapons_F", // for overwriting V_RebreatherB with attribute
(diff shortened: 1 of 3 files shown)


addons/core/config.cpp
```diff
@@ -17,6 +17,7 @@ class CfgPatches {
             "cba_ui",
             "cba_xeh",
             "cba_settings",
+            "A3_Weapons_F", // for overwriting V_RebreatherB with attribute
```
dedmen, line 20: not needed as CBA pulls in all basegame addons, but also doesn't hurt

Issue #1463, opened 2019-04-25
Change TS3 nicknames from profile name to steamID 64


sp9dev:
Is it possible to set up TFAR in a way, that instead of changing TS3 name to Arma 3 Profile Name, it would change player's TS3 nickname to thair steamID64?

dedmen:
Set the players ingame name to that.

vabene1111:
@sp9dev  it is also relativly easy to find all places that `name` or `profileName` is used and replace them by something else (just search trough all files for it).

I dont know if looking at the teamspeak channel for information on whom you are talking to/who is in the vicinity is a relevant problem for the milsim community (which TFAR is primarily intended for as far as i know), i guess its mostly a problem for everythign RPG based. 

If @dedmen approves i would try to open a pull request to handle the naming over a variable on the player and therfore allow customization trough this ?

dedmen:
I wanted to decouple TS and ingame names anyway.
There is no technical limitation that requires having same names, just a human worktime limitation.

Issue #1168, opened 2016-11-30
Feature request - don't automatically rename players on teamspeak


Reekoj:
Hello.

I was wondering if you could consider adding the ability to enable/disable an option in a configuration file, that would stop renaming players on teamspeak when they join the server

dedmen:
Renaming is an essential part.. Actually its the only way we identify players. It would also be possible to let the Plugin ask every user in the channel for their ingame name. This would potentially increase the Bandwidth load on the Teamspeak server by a lot. It would add more Logic to care for performance and maintenance wise. It is possible with the new Plugin structure in 1.0. But I am against this. Also thought some times about doing that but the Pro's are not enough to overweigh the Cons.

DuckyDoesCode:
I would like this feature as on my community we use guids as the ingame player tags and currently its hard to be able to tell who is who and it would be nice of having a script to call to change their teamspeak name to their guid instead.

dedmen:
Use SteamUID as player identifier instead.
This would also make spectator code easier because it has to get the playername which it can't when player is dead and spectating.

PR #1496, opened 2020-01-06
added ability to set custom teamspeak name

vabene1111:
This addresses the ideas/problems mentioned in #1463 and #1168 by allowing the name shown in TeamSpeak (and therefore used by the plugin) to be dynamically set using the variable `TFAR_unitName`.

- Modify all scripts using the players name to use one stored in a variable instead
- not change anything for players/developers if they don't use this feature

There are 2 considerations that need to be decided before merging this:
- should `TFAR_unitName` be set in some initialization script so defaults must not be used everywhere for performance reasons ? I don't know if this matters.
- can `TFAR_spectatorName` and `TFAR_unitName` be merged in one variable ? i don't see where it is ever set to something different than the players name but i don't know if there is some technical reason.

Using a variable instead of the name directly has been tested on a large scale for the past two years on our server and has so far not revealed any problems.

### DIFF
--- a/addons/core/functions/fnc_forceSpectator.sqf
+++ b/addons/core/functions/fnc_forceSpectator.sqf
@@ -25,7 +25,7 @@ params ["_player", "_value"];
 _player call TFAR_fnc_releaseAllTangents; //Stop all radio transmissions
 
 _player setVariable ["TFAR_forceSpectator", _value, true];
-_player setVariable ["TFAR_spectatorName", profileName, true];
+_player setVariable ["TFAR_spectatorName", (TFAR_currentUnit getVariable ["TFAR_unitName", profileName]), true];
 
 if (TFAR_ShowVolumeHUD && _value) then { //Don't have a voice in spectator Mode so hide the HUD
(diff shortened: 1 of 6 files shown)


Reekoj:
@vabene1111 are you still active ? I think this pull request is not a draft anymore

vabene1111:
It's been a long time since I worked on this but the feature has also been acimtivemy used all that time and is well tested. Basically I think I only set it to draft back then to decide what to do with the variables.

You can either just quickly make that change after  merging or I can look into it, whatever you want.

vabene1111:
Cannot remove draft on mobile so either you can or I will when I am back on my PC

Reekoj:
I have no rights on the repo i can't do anything lol, but Dedmen said he can't merge it because it's in draft status

vabene1111:
ah ok, yea found it on GH pc version website, now it should be mergable. The code still looks fine and the change with the variables can always easily be added afterwards so i think this is ready to merge

dedmen:
Thank you!

PR #1505, opened 2020-05-21
Updated the ACE interaction radio icon

Avokadomos:
Created a new icon to replace the old one, with a visual style matching those of the ACE interaction wheel.

image of radio icons


dedmen:
But I liked my old icon 😢

PR #1508, opened 2020-08-01
Added - Check for vehicleIDOverride existing on unit

Saborknight:
Why
=================
To allow for intercom use outside of a vehicle, but linked to an
intercom of a vehicle or set of vehicles.

Issue ID: N/A

- Allow units to use the intercom whenever the vehicleIDOverride variable is applied to them
- Intended for behind the scenes use, which is why no 3DEN attributes have been created to accompany this

### DIFF
--- a/addons/core/functions/fnc_vehicleId.sqf
+++ b/addons/core/functions/fnc_vehicleId.sqf
@@ -19,7 +19,7 @@
 */
 params ["_unit"];
 
-if (isNull (objectParent _unit) && isNil _unit getVariable "TFAR_vehicleIDOverride") exitWith {"no"};//Unit is not in vehicle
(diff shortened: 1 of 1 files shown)


addons/core/functions/fnc_vehicleId.sqf
```diff
@@ -19,7 +19,7 @@
 */
 params ["_unit"];
 
-if (isNull (objectParent _unit)) exitWith {"no"};//Unit is not in vehicle
+if (isNull (objectParent _unit) && isNil _unit getVariable "TFAR_vehicleIDOverride") exitWith {"no"};//Unit is not in vehicle
```
shukari, line 22:
```suggestion
if (isNull (objectParent _unit) && isNil {_unit getVariable "TFAR_vehicleIDOverride"}) exitWith {"no"};//Unit is not in vehicle
```

PR #1517, opened 2020-11-14
GitHub Actions

Dahlgren:
- Add HEMTT project specification
- Build Arma 3 addons using HEMTT (no binarization of p3ds)
- Build Arma 3 extension, 32 bit + 64 bit
- Build Teamspeak plugin, 32 bit + 64 bit
- Upload archive with everything packed together

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,164 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Git checkout
+        uses: actions/checkout@v2
+
+      - name: Validate config
+        run: python3 tools/validator_cfg.py
+
+      - name: Validate SQF
+        run: python3 tools/validator_sqf.py
+
+      - name: Validate XML
+        run: python3 tools/validator_xml.py
+
+  mod:
+    name: Mod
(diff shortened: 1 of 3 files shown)


Dahlgren:
Example build, https://github.com/Dahlgren/task-force-arma-3-radio/runs/1400352630

jokoho48:
why not use the offical hemtt github action? https://github.com/arma-actions/hemtt

Dahlgren:
It (currently) cannot be locked to specific HEMTT version and new releases are not always backwards compatible. For example the 0.7 versions do not support json project files anymore. Using a specific version avoids suddenly breaking the TFAR builds.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,164 @@
...
+      - name: Git checkout
+        uses: actions/checkout@v2
+        with:
+          lfs: true
+
+      - name: HEMTT Build
+        uses: docker://hemtt/hemtt:0.7.6
+        with:
+          entrypoint: hemtt
+          args: build --release
```
Dahlgren, line 35: `--ci` should be added here for less fancy terminal output

dedmen:
No need, appveyor does it all correctly.
And this is missing stuff like dll signing, correct build versions, pdb upload for dll's (which can be fixed)
Also how does hemtt binarize p3d's without Arma tools being available?
Also you cannot rebuild the Arma extensions. BattlEye.

PR #1530, opened 2021-06-23
External Intercom

Saborknight:
- Let people ACE Interact with a vehicle to pickup/put away an intercom phone, allowing them to speak on the intercom of any ground vehicle
- Let people ACE Interact with a vehicle to connect wirelessly to the vehicle's intercom, allowing them to speak on the intercom of any ground or air vehicle, assuming that they're wearing compatible headgear
- Whenever a player wearing a wireless compatible headgear jumps out of an intercom capable vehicle's non-cargo seat, they are automatically connected to the intercom wirelessly, so that they can continue talking on the intercom without having to interact
- In all above cases, each of the two methods have their distinct maximum range of operation (Phone: 5m, Wireless: 15m). These values are customisable via CBA Settings

~~I'll be adding to this pull request with any suggested edits made, and by covering the below currently un-handled situations.~~ I think I'm done in terms of adding new features. ~~I'll also be making another pull request which will implement the physical telephone and rope, to add to immersion.~~ So it turns out I've done everything in one pull request. Oh well.

**Situations to handle correctly**
- [x] Player moves outside of the 5 meter limit -> Intercom phone disconnected
- [x] Another player tries to pickup an intercom phone whilst someone else has already picked up the same phone -> Display an "Intercom phone is in use" ACE Interaction instead of the "Pickup Intercom Phone" ACE interaction. Also, Intercom phone will not activate if already in use, in case the ACE interaction is bypassed
- [x] Player is killed, or disconnects from the server whilst connected to the intercom phone -> Intercom phone disconnected
- [x] Player gets into the same vehicle they are connected to with the intercom phone -> Intercom phone disconnected
- [x] Player gets into *any* vehicle whilst they are connected to the intercom phone -> will disconnect, though the distance checks should come into play if the vehicle the player is in, or the vehicle they are connected to move outside of the 5 meter radius
- [x] Player picks up another intercom phone whilst still being connected to a different intercom phone -> will disconnect old intercom phone
- [x] Player going unconscious (ACE) whilst being connected to the intercom phone -> will disconnect

**Todo for this initial implementation (AKA 1.0?)**
- [x] Add: option to disable external intercom, or only same side can connect
- [x] Add: Have wireless headgear reconnect automatically in case someone runs out of range by mistake. Make it so that they must disconnect via self interaction.
- [x] Allow wireless users to change intercom channel
- [x] Add: channel switch to phone panel, not self interact
- [x] Add: attribute that can be added to individual vehicle from Eden, Config or Zeus(?) to customise max range of phone and/or wireless
- [x] Fix: When jumping out of a vehicle, the intercom does stay connected wirelessly, but the icon doesn't appear?
- [ ] Fix: Rope breaking doesn't disconnect
- [x] Fix: AI jumping out will show indicator icon, even if Zeus is not on intercom
- [x] Fix: Now Headgear whitelist is suddenly not being transferred from server to client anymore...?
- [x] When being connected to wireless intercom, and changing the "Enable" setting to "Same side as Player", ropes attach to all players currently connected to the wireless intercom. So, one of the 2 players was given the rope and phone, but not actually assigned as phone operator... Meaning that player now can't put away the attached rope and phone
- [x] When one wireless user disconnects, all players lose the wireless indicator. This does not apply to someone when jumping in the vehicle (and wireless intercom disonnecting normally for them)
- [x] When a phone user is connected, and a crew member jumps out of the vehicle, the phone and cable disappear but icon remains, and intercom is still connected
- [x] When a wireless user takes helmet off, whilst another user is connected to wireless, it removes the phone and cable, but icon remains and intercom stays connected
- [x] Changing server settings don't apply globally
- [x] When no users are connected to the phone, but all are connected to the wireless, and the setting Enable is changed to something, then a random user (I assume first in the list) is actually connected to the phone instead of the wireless, and the other users are disconnected completely
- [x] When 2 users connected wirelessly, and setting gone from 1 to 2, locally each user is still considered connected, but the `TFAR_ExternalIntercomSpeakers` variable of the vehicle is not syncronised so we're each seeing an option to disconnect ourselves
- [x] When in crew seat, change to crew, then dismount. The channel will not remain on crew, it'll switch back to default
- [x] Connecting to wireless, getting into crew position, then getting out will duplicate person entry in `TFAR_ExternalIntercomSpeakers` of the vehicle
- [x] A player who isn't connected to intercom at all will have the icon appear randomly when another player jumps out of the vehicle
- [x] Modded Vehicles don't have the Intercom Phone interaction points?
- [x] Add entry to ACE Arsenal to indicate if a headgear is External Intercom Compatible
- [ ] ~~When all users are connnected to the wireless and the Enable setting is changed to 2, the second player in the list still has the "Disconnect Wireless" ACE interaction showing on the vehicle, even though they've been otherwise completely disconnected~~
- [ ] ~~A player (previously connected to wireless) isn't disconnected from the intercom when disconnecting from the server~~
- [ ] ~~Fix: Make sure `fnc_getIntercomChannel` sorts the array by `commander` > `gunner` > `driver` > `passengers` instead of the current sorting~~ Not so important if we just give the ability for the person to change channel themselves

> Note: Credits to Gary Sinise for donation of the handset model as well as corrections to certain calculations involved (He's the Maths voodoo man)


addons/intercom_phone/ACEActions.hpp
```diff
@@ -0,0 +1,31 @@
+#define ADD_ACTIONS(parent, child) \
+class child : parent { \
+    class ACE_Actions { \
+        class ACE_MainActions { \
+            class TFAR_Intercom_Phone_Activate { \
+                displayName = CSTRING(PICKUP); \
+                distance = 4; \
+                condition = "isNil {_player getVariable ""TFAR_vehicleIDOverride""} && ([(typeOf _target), ""TFAR_hasIntercom"", 0] call TFAR_fnc_getVehicleConfigProperty) > 0 && isNil {_target getVariable [""TFAR_IntercomPhoneSpeaker"", nil]}"; \
```
dedmen, line 8:
```suggestion
                condition = "isNil {_player getVariable ""TFAR_vehicleIDOverride""} && ([(typeOf _target), ""TFAR_hasIntercom"", 0] call TFAR_fnc_getVehicleConfigProperty) > 0 && isNil {_target getVariable ""TFAR_IntercomPhoneSpeaker""}"; \
```

addons/intercom_phone/CfgVehicles.hpp
```diff
@@ -0,0 +1,16 @@
+class CfgVehicles {
+    #include "ACEActions.hpp"
+
+    class LandVehicle;
+    ADD_ACTIONS(LandVehicle, Car)
+    ADD_ACTIONS(LandVehicle, Tank)
+    ADD_ACTIONS(LandVehicle, Motorcycle)
```
dedmen, line 7: Do Motorcycle's ever have intercoms? :D 

I'd say that can be skipped, more important here would be helicopters I'd say

addons/intercom_phone/RscTitles.hpp
```diff
@@ -0,0 +1,28 @@
...
+            class ConnectionIndicator: RscPictureKeepAspect {
+                idc= 1112;
+                type = 0;
+                style = "0x30 + 0x800";
+                colorText[] = { 1, 1, 1, 1 };
+                colorBackground[]={0, 0, 0, 0};
+                font = "PuristaMedium";
+                sizeEx = "(((((safezoneW / safezoneH) min 1.2) / 1.2) / 25) * 0.8)";
+                text=QPATHTOF(ui\tfar_intercom_phone.paa);
+                x="(profilenamespace getvariable [""IGUI_grid_TFAR_Volume_X"",  0.85 * safezoneW + safezoneX])";
```
dedmen, line 21: Nice idea to reuse this.
But people can enable volume indicator to always show, that means these will then overlap eachother?
Maybe want to hide the speech volume indicator while you're in intercom

dedmen: Actually you added a grid for intercom phone in config, but here you seem to use the volume one

addons/intercom_phone/functions/fnc_activate.sqf
```diff
@@ -0,0 +1,73 @@
...
+
+private _phoneSpeaker = _vehicle getVariable ["TFAR_IntercomPhoneSpeaker", nil];
+
+if (!isNil "_phoneSpeaker") exitWith {
+    // [parseText format["<br /><br /><t align='center'>%1 is currently on the phone.</t>", name _phoneSpeaker]] call TFAR_fnc_showHint;
+    "no";
+};
+
+private _vehicleID = _vehicle getVariable ["TFAR_vehicleIDOverride", netid _vehicle];
+_vehicle setVariable ["TFAR_vehicleIDOverride", _vehicleID, true];
```
dedmen, line 33: Is this needed? considering you just read that variable and know that (atleast locally) it already has that value?

addons/intercom_phone/functions/fnc_activate.sqf
```diff
@@ -0,0 +1,73 @@
...
+_player setVariable ["TFAR_IntercomPhoneVehicle", _vehicle, true];
+_player setVariable ["TFAR_vehicleIDOverride", _vehicleID, true];
+
+// Get intercom channel, now that we've signed on to the intercom
+private _intercomSlot = [_vehicle] call TFAR_intercom_phone_fnc_getIntercomChannel;
+_vehicle setVariable [format ["TFAR_IntercomSlot_%1",(netID _player)], _intercomSlot, true];
+_player setVariable [format ["TFAR_IntercomSlot_%1",(netID _player)], _intercomSlot, true]; // Because TFAR checks the "vehicle", and thinks the unit is the vehicle when it's not in a vehicle, so we manually set it
+
+// Show indicator
+if (TFAR_oldVolumeHint) then {
```
dedmen, line 44: Here we want to hide the volume indicator

addons/intercom_phone/functions/fnc_activate.sqf
```diff
@@ -0,0 +1,73 @@
...
+    // [parseText "Connected to Vehicle Intercom", -1] call TFAR_fnc_showHint;
+} else {
+    (QGVAR(ConnectionIndicatorRsc) call BIS_fnc_rscLayer) cutRsc [QGVAR(ConnectionIndicatorRsc), "PLAIN", 0, true];
+};
+
+private _loopID = [_vehicle, _player] spawn {
+    params ["_vehicle", "_player"];
+
+    while {true} do {
```
dedmen, line 53: I don't like spawned scripts. And also not while true loops. But this is appropriate here I guess.

1. The while loop should exit after deactivate is called
2. boundingBoxReal doesn't need to be re-calculated every time, that can be done once outside of the loop.

addons/intercom_phone/functions/fnc_busyInteraction.sqf
```diff
@@ -0,0 +1,32 @@
+#include "script_component.hpp"
+
+/*
+  Name: TFAR_fnc_intercomBusyIntercation
```
dedmen, line 4:
```suggestion
  Name: TFAR_fnc_intercomBusyInteraction
```

addons/intercom_phone/functions/fnc_getIntercomChannel.sqf
```diff
@@ -0,0 +1,28 @@
...
+    [_vehicle, _player] call TFAR_fnc_getIntercomChannel;
+
+  Public: Yes
+*/
+
+params ["_vehicle"];
+
+// Get any occupants, from crew down to passengers
+private _crew = crew _vehicle;
+_vehicle getVariable [format ["TFAR_IntercomSlot_%1",(netId (_crew select 0))],
```
dedmen, line 26: Is this variable set locally? I thought it was local only to remember what channel the player was last in when he left the vehicle.

addons/intercom_phone/functions/fnc_activate.sqf
jokoho48: wouldnt something like this work:
```sqf
private _bbSize = (((boundingBoxReal _vehicle) select 2) / 2) + 5;
[{
    params ["_vehicle", "_player", "_bbSize"];
    (_vehicle distance _player) <= _bbSize
}, {
    params ["_vehicle", "_player"];
    [_vehicle, _player] call TFAR_intercom_phone_fnc_deactivate;
}, [_vehicle, _player, _bbSize]] call CBA_fnc_waitUntilAndExecute
```

addons/intercom_phone/RscTitles.hpp
Saborknight: Honestly, I don't understand dialogues and, more specifically, how to position them at all... So was hoping someone would come round and comment on it like you did. So this grid, do I use the class name specified in https://github.com/michail-nikolaev/task-force-arma-3-radio/pull/1530/files#diff-5238f9c6d8630cb9ec6d642487487911310e6619735d338c5ea33bf702fa4aa2R46 ?

addons/intercom_phone/functions/fnc_activate.sqf
Saborknight: Thanks for that @jokoho48 ! I'll use your suggestion instead. It did a couple of errors in it but it ultimately works fine.

addons/intercom_phone/functions/fnc_getIntercomChannel.sqf
```diff
@@ -0,0 +1,28 @@
...
+  Example:
+    [_vehicle, _player] call TFAR_fnc_getIntercomChannel;
+
+  Public: Yes
+*/
+
+params ["_vehicle"];
+
+// Get any occupants, from crew down to passengers
+private _crew = crew _vehicle;
```
Saborknight, line 25: It actually sucks that `crew` actually returns the list sorted by driver, gunner, commander and then passengers... when in actual fact it makes more sense to connect to the commander's channel, then the gunner's, then the driver's, then the passengers if the previous are not present...

addons/intercom_phone/functions/fnc_getIntercomChannel.sqf
dedmen: > Do you mean that the variable for that crew member is local to that crew member's machine only?

Yes, atleast it doesn't make any sense for other players to know that, its only used for local cache afaik

addons/external_intercom/baseHeadgear.hpp
```diff
@@ -0,0 +1,20 @@
+private _baseHeadgear = [
```
dedmen, line 1: This should be a config property, what should mods do to add their supported helmets to this?
I'd say make it config property, and scan once at preStart and save it in uiNamespace.

Or alternatively instead of scanning, could also make a global config class/array that mods can append to

addons/external_intercom/functions/fnc_connect.sqf
```diff
@@ -0,0 +1,148 @@
...
+_vehicle setVariable ["TFAR_ExternalIntercomSpeakers", _externalSpeakers, true];
+
+// Get intercom channel, now that we've signed on to the intercom
+private _intercomSlot = [_vehicle] call TFAR_external_intercom_fnc_getIntercomChannel;
+
+_vehicle setVariable [format ["TFAR_IntercomSlot_%1",(netId _player)], _intercomSlot, true];
+_player setVariable [format ["TFAR_IntercomSlot_%1",(netId _player)], _intercomSlot, true]; // Because TFAR checks the "vehicle", and thinks the unit is the vehicle when it's not in a vehicle, so we manually set it
+
+// Show indicator
+if (TFAR_oldVolumeHint) then {
```
dedmen, line 65: You don't need to use this. This was only specifically for Volume indicator (as some people didn't want the new on-screen indicator).

This is a new feature, you don't need to worry about backwards compat and people "preferring the old way"
I'd say take that out and only ever use the new indicator.

Or if you want to keep it, its unrelated to volume display so it should probably be separate setting then

addons/external_intercom/functions/fnc_connect.sqf
```diff
@@ -0,0 +1,148 @@
...
+
+
+if !(_wireless) then {
+    private _handset = createSimpleObject [QPATHTOF(data\TFAR_handset.p3d), _player selectionPosition "head"];
+    _handset attachTo [_player, [-0.14,-0.02,0.02], "head", true];
+    _handset setVectorDirAndUp [[-2.5,0.8,0.25],[-1,-1,1]];
+    private _ropeID = ropeCreate [_vehicle, _interactionPointRelative vectorAdd [0,0.2,0], _handset, "plug", _intercomMaxRange]; // handset rope location[0,0.006,-0.7]
+    _vehicle setVariable ["TFAR_ExternalIntercomRopeIDs", [_ropeID, _handset], true];
+
+    _vehicle addEventHandler ["RopeBreak", {
```
dedmen, line 101: Hah. This alone might be a sufficient "out of range" check :D

addons/external_intercom/functions/fnc_connect.sqf
```diff
@@ -0,0 +1,148 @@
...
+private _headgear = "";
+
+if (!_wireless) then {
+    _interactionPointRelative = [_vehicle] call TFAR_external_intercom_fnc_getInteractionPoint;
+    _intercomMaxRange = TFAR_externalIntercomMaxRange_Phone;
+} else {
+    _headgear = headgear _player;
+};
+
+[{
```
dedmen, line 83: I'd suggest optimizing this by writing two conditions, one for wireless and one for not wireless.

You don't need modelToWorld (somewhat expensive) if its wireless, as center of mass is usually 0,0,0 so you can just use vehicle pos.

You also don't need to do the wireless check if not wireless.

You can do one check here, to save performance on every frame

addons/external_intercom/functions/fnc_connect.sqf
```diff
@@ -0,0 +1,148 @@
...
+        }],
+        true
+    ];
+};
+
+_player setVariable ["TFAR_ExternalIntercomEHID",
+    ["ace_unconscious", {
+        params ["_player", "_active"];
+        if (_active) then {
+            [_player getVariable "TFAR_ExternalIntercomVehicle", _player] call TFAR_external_intercom_fnc_disconnect;
```
dedmen, line 145: Should it really disconnect if you go unconscious though?
If its wireless you still have your helmet/headset on even if you go unconscious

addons/external_intercom/functions/fnc_connect.sqf
```diff
@@ -0,0 +1,148 @@
...
+            } forEach ((_vehicle getVariable ["TFAR_ExternalIntercomSpeakers", [objNull, []]]) select 1);
+            _vehicle removeMPEventHandler ["MPKilled", _thisEventHandler];
+        }],
+        true
+    ];
+};
+
+_player setVariable ["TFAR_ExternalIntercomEHID",
+    ["ace_unconscious", {
+        params ["_player", "_active"];
```
dedmen, line 143: should check player? Afaik ace_unconscious can also fire for local AI?

addons/external_intercom/functions/fnc_disconnect.sqf
```diff
@@ -0,0 +1,50 @@
...
+_player setVariable ["TFAR_ExternalIntercomVehicle", nil, true];
+
+_externalIntercomSpeakers = _vehicle getVariable ["TFAR_ExternalIntercomSpeakers", [objNull, []]];
+
+if ((_externalIntercomSpeakers select 0) isEqualTo _player) then {
+    _externalIntercomSpeakers set [0, objNull];
+};
+(_externalIntercomSpeakers select 1) deleteAt ((_externalIntercomSpeakers select 1) find _player);
+
+_vehicle setVariable ["TFAR_ExternalIntercomSpeakers", _externalIntercomSpeakers, true];
```
dedmen, line 35: potential race condition if two players disconnect at the same time in MP.
player1 might re-add player2 to the array even after player2 disconnected.
Not sure how to solve this, and also probably not that significant

addons/external_intercom/functions/fnc_getInteractionPoint.sqf
```diff
@@ -0,0 +1,33 @@
+params ["_vehicle"];
+
+private _customPosition = getArray(configOf _vehicle);
```
dedmen, line 3: does this make sense?
How can you call getArray on a vehicle class??

addons/external_intercom/functions/fnc_getInteractionPoint.sqf
```diff
@@ -0,0 +1,33 @@
...
+
+// If there's a custom position, then don't bother calculating
+if !(_customPosition isEqualTo []) exitWith {
+   _customPosition;
+};
+
+private [
```
dedmen, line 10: Don't use private array, its a performance waste.
use the private keyword

addons/external_intercom/functions/fnc_parseClassnamesInput.sqf
```diff
@@ -0,0 +1,49 @@
...
+};
+
+_valueString = _valueString splitString " " joinString ""; //remove spaces
+private _parsedValue = (parseSimpleArray _valueString) apply {
+    if !(IS_STRING(_x)) then {format["%1", _x]} else {_x};
+};
+
+_parsedValue = _parsedValue arrayIntersect _parsedValue; // Retain only unique elements
+
+_parsedValue = _parsedValue apply {
```
dedmen, line 45: Important! You are using `in` to check if headgear matches. `in` is case sensitive.
Meaning here you are relying on players to enter case-sensitively correct classnames. That is unlikely.


Instead of returning _x below, return `configName(configFile >> "CfgWeapons" >> _x)` that will get you correct configCase

addons/external_intercom/functions/fnc_connect.sqf
Saborknight: Oh hell no, I prefer the icon interface by far

addons/external_intercom/functions/fnc_connect.sqf
Saborknight: I wish it was. The rope will stretch for ages rather than snap... which is kind of annoying. I wish we could set the elasticity of the rope to tell the engine how much give it should allow.

Actually... Does it exist?

addons/external_intercom/functions/fnc_connect.sqf
Saborknight: Good point

addons/external_intercom/functions/fnc_connect.sqf
Saborknight: I didn't think about that, good point

addons/external_intercom/functions/fnc_getInteractionPoint.sqf
Saborknight: Whoops

addons/external_intercom/functions/fnc_connect.sqf
Saborknight: Once v2.06 of Arma is released we'll be able to make a custom rope to imitate wire, and change elasticity accordingly.

The code:
```c
class CfgVehicles {
    class Rope;
    class RopeRabbit : Rope {
        maxRelLenght = 1.1;
        maxExtraLenght = 20;
        springFactor = 0.5; // higher == less stretchy rope
        segmentType = "RopeSegmentRabbit";
        torqueFactor = 0.5;
        dampingFactor[] = {1.0,2.5,1.0};
        model = "\A3\Animals_F\rabbit\rabbit_F.p3d";
    };
};

class CfgNonAIVehicles {
    class RopeSegmentRabbit {
        scope=2;
        displayName="";
        simulation="ropesegment";
        autocenter=0;
        animated=0;
        model = "\A3\Animals_F\rabbit\rabbit_F.p3d";
    };
};
```
And to create it, just name the classname as an additional parameter to the usual ones for `ropeCreate`

addons/intercom_phone/functions/fnc_getIntercomChannel.sqf
Saborknight: Eh.... Let me get back to you on this after some testing with my guys.

addons/external_intercom/functions/fnc_disconnect.sqf
Saborknight: I'm not sure how to deal with the possible race condition either... maybe someone else will have an epiphane for us.

Saborknight:
- [ ] Add option to disable this, or only same side can connect
- [ ] Idea: Have wireless headgear reconnect automatically in case someone runs out of range by mistake. Make it so that they must disconnect via self interaction.
Also, allow wireless users to change intercom channel
Add channel switch to phone panel, not self interact
Add attribute that can be added to individual vehicle from Eden, Config or Zeus(?) to customise max range of phone and/or wireless
Issue: When jumping out of a vehicle, the intercom does stay connected wirelessly, but the icon doesn't appear?
Issue: Rope breaking doesn't disconnect

Saborknight:
Some change since the last review has caused the headgear setting not to be transferred to clients anymore? When I fetch the setting from the server, it looks the way it should, but when I just try and use the global variable normally via just stating it, it comes up blank. I was under the impression that CBA would handle the transfer of the variable to any clients with the `onChange` parameter.... but it looks like I need to do it manually...

addons/external_intercom/functions/fnc_disconnect.sqf
dedmen: Let the server handle it.
Like with radio ID's. Tell the server "hey that happened" and the server takes care of checking and doing.
The server cannot do two things in parallel.

addons/core/functions/fnc_preparePositionCoordinates.sqf
```diff
@@ -27,9 +27,13 @@ private _isSpectating = _unit getVariable ["TFAR_forceSpectator",false];
 
 private _isolated_and_inside = false; //_isInVehicle && {_unit call TFAR_fnc_vehicleIsIsolatedAndInside};
 private _vehicle = "no"; //if (_isInVehicle) then {_unit call TFAR_fnc_vehicleId} else {"no"};
-if (!isNull (objectParent _unit) && !_isSpectating) then {//_isInVehicle
+private _isInVehicle = !isNull (objectParent _unit);
+if ((!(isNil {_unit getVariable "TFAR_vehicleIDOverride"}) || _isInVehicle) && !_isSpectating) then {//_isInVehicle
```
dedmen, line 31: check _isInVehicle first, and do the isNil in lazy eval, isNil is somewhat a bit more expensive.

Maybe its cheaper to not use isNil?
maybe default value something like ctrlNull and then isEqualTo ctrlNull.
Would need to benchmark that.

This function runs for every unit for (almost) every frame, it needs to be top performing

addons/core/functions/fnc_vehicleId.sqf
```diff
@@ -19,7 +19,9 @@
 */
 params ["_unit"];
 
-if (isNull (objectParent _unit) && isNil {_unit getVariable "TFAR_vehicleIDOverride"}) exitWith {"no"};//Unit is not in vehicle
+private _vehicleOverrideEnabled = !isNil {_unit getVariable "TFAR_vehicleIDOverride"};
```
dedmen, line 22: this is the hot path.
The normal case is that its not enabled and the function immediately exits returning "no"

you invert the value with `!` twice in the hotpath, if you remove the ! on both you save two command calls

addons/external_intercom/CustomInteractionPoints.hpp
```diff
@@ -0,0 +1,97 @@
...
+};
+
+class Tank_F;
+class MBT_04_base_F : Tank_F {
+    TFAR_ExternalIntercomInteractionPoint[] = {-1.44043,-5.65282,-1.23575};
+};
+
+class Tank_F;
+class MBT_02_base_F : Tank_F {
+    TFAR_ExternalIntercomInteractionPoint[] = {-1.26074,-4.75584,-0.938046};
```
dedmen, line 75: would be cool if this can also support string memory point

addons/external_intercom/WirelessHeadgear.hpp
```diff
@@ -0,0 +1,19 @@
+class TFAR_External_Intercom_Wireless_Headgear {
+    // Vanilla + Apex + Contact
+    class H_HelmetCrew_B;
```
dedmen, line 3: What are these classes here?
just use entry=1 like Joint Rails and some things in Arma do?

addons/external_intercom/XEH_preInit.sqf
```diff
@@ -0,0 +1,61 @@
...
+[
+    "TFAR_setting_externalIntercomWirelessHeadgear",
+    "EDITBOX",
+    [LSTRING(SETTING_WIRELESS_HEADGEAR_HEADER), LSTRING(SETTING_WIRELESS_HEADGEAR_DESC)],
+    [localize ELSTRING(settings,global), localize LSTRING(SETTINGS_CATEGORY_EXTERNAL_INTERCOM)],
+    ["", false, FUNC(settingForceAlphaNumericArray)],
+    1,
+    {
+        if (isServer) then {
+            TFAR_externalIntercomWirelessHeadgear = ((configFile >> "TFAR_External_Intercom_Wireless_Headgear") call BIS_fnc_getCfgSubClasses) + ([_this] call FUNC(parseClassnamesInput));
```
dedmen, line 14: just directly https://community.bistudio.com/wiki/configProperties with "true" condition might be cheaper than calling the BIS func

addons/external_intercom/XEH_preInit.sqf
```diff
@@ -0,0 +1,61 @@
...
+[
+    "TFAR_externalIntercomEnable",
+    "LIST",
+    [LSTRING(SETTING_ENABLE_FOR_HEADER), LSTRING(SETTING_ENABLE_FOR_DESC)],
+    [localize ELSTRING(settings,global), localize LSTRING(SETTINGS_CATEGORY_EXTERNAL_INTERCOM)],
+    [[0, 1, 2], [LSTRING(SETTING_ENABLE_0), LSTRING(SETTING_ENABLE_1), LSTRING(SETTING_ENABLE_2)], 0],
+    1,
+    {
+        _vehicle = ACE_Player getVariable "TFAR_ExternalIntercomVehicle";
+        if (_this > 0 && !isNil "_vehicle") then {
```
dedmen, line 29: But if you disable it mid game, this doesn't disconnect you?

addons/external_intercom/data/model.cfg
```diff
@@ -0,0 +1,25 @@
+class CfgSkeletons
+{
+	class Default
+	{
+		isDiscrete=1;
```
dedmen, line 5: 4 space indent. space before and after `=`

addons/external_intercom/functions/fnc_addIntercomChannels.sqf
```diff
@@ -0,0 +1,89 @@
...
+
+_intercomChannels = _intercomChannels apply {
+    _statement = getText(_intercomRootClass >> _x >> "statement");
+    // Since we're dealing with unpredictable developers, search a string of 5
+    // characters then filter out anything non-numeric but include negative (-)
+    _channelId = parseNumber ([_statement select [((_statement find _searchTerm) + count _searchTerm),5], "-0123456789"] call BIS_fnc_filterString);
+    [getText(_intercomRootClass >> _x >> "displayName"), _channelId];
+};
+
+if (_intercomChannels isEqualTo []) exitWith {false;};
```
dedmen, line 40:
```suggestion
if (_intercomChannels isEqualTo []) exitWith {false};
```

addons/external_intercom/functions/fnc_addIntercomChannels.sqf
```diff
@@ -0,0 +1,89 @@
...
+
+private _actions = [];
+
+{
+    _x params ["_displayName", "_intercomChannel"];
+
+    _actions pushBack [
+        [
+            format["TFAR_ExternalIntercomChannel_%1", _intercomChannel],
```
dedmen, line 49:
```suggestion
            format ["TFAR_ExternalIntercomChannel_%1", _intercomChannel],
```

addons/external_intercom/functions/fnc_addIntercomChannels.sqf
```diff
@@ -0,0 +1,89 @@
...
+            _displayName,
+            "",
+            {
+                params ["_vehicle", "_player", "_params"];
+                _params params ["_intercomChannel", "_displayName"];
+
+                if (!isNil "_vehicle") then {
+                    _vehicle setVariable [format ["TFAR_IntercomSlot_%1",(netId _player)], _intercomChannel, true];
+                    _player setVariable [format ["TFAR_IntercomSlot_%1",(netId _player)], _intercomChannel, true];
```
dedmen, line 58:
```suggestion
                    _vehicle setVariable [format ["TFAR_IntercomSlot_%1", (netId _player)], _intercomChannel, true];
                    _player setVariable [format ["TFAR_IntercomSlot_%1", (netId _player)], _intercomChannel, true];
```

addons/external_intercom/functions/fnc_addIntercomChannels.sqf
```diff
@@ -0,0 +1,89 @@
...
+            {},
+            [_intercomChannel, _displayName, _vehicle],
+            {[0,0,0]},
+            4,
+            [false,false,false,false,false],
+            {
+                params ["_vehicle", "_player", "_args", "_actionData"];
+                _args params ["_intercomChannel", "_displayName"];
+
+                _intercom = _player getVariable [format ['TFAR_IntercomSlot_%1',(netID _player)], -2];
```
dedmen, line 71:
```suggestion
                _intercom = _player getVariable [format ['TFAR_IntercomSlot_%1', (netID _player)], -2];
```

addons/external_intercom/functions/fnc_addWirelessIntercomMenu.sqf
```diff
@@ -0,0 +1,67 @@
...
+
+  Public: Yes
+*/
+params ["_target", "_player"];
+
+_vehicle = _player getVariable "TFAR_ExternalIntercomVehicle";
+
+if (isNil "_vehicle") exitWith {
+    [];
+};
```
dedmen, line 27:
```suggestion
if (isNil "_vehicle") exitWith {[]};
```

addons/external_intercom/functions/fnc_addWirelessIntercomMenu.sqf
```diff
@@ -0,0 +1,67 @@
...
+    [
+        format["TFAR_External_Intercom_Wireless_Disconnect", _intercomChannel],
+        localize LSTRING(DISCONNECT_WIRELESS),
+        QPATHTOF(ui\tfar_ace_interaction_external_intercom_wireless_disconnect.paa),
+        {
+            params ["_vehicle", "_player"];
+            [_player getVariable 'TFAR_ExternalIntercomVehicle', _player] call TFAR_external_intercom_fnc_disconnect;
+        },
+        {
+            params ["_vehicle", "_player"];
```
dedmen, line 41:
```suggestion
            params ["", "_player"];
```

addons/external_intercom/functions/fnc_addWirelessIntercomMenu.sqf
```diff
@@ -0,0 +1,67 @@
...
+
+private _actions = [];
+
+_actions pushBack [
+    [
+        format["TFAR_External_Intercom_Wireless_Disconnect", _intercomChannel],
+        localize LSTRING(DISCONNECT_WIRELESS),
+        QPATHTOF(ui\tfar_ace_interaction_external_intercom_wireless_disconnect.paa),
+        {
+            params ["_vehicle", "_player"];
```
dedmen, line 37:
```suggestion
            params ["", "_player"];
```

addons/external_intercom/functions/fnc_addWirelessIntercomMenu.sqf
```diff
@@ -0,0 +1,67 @@
...
+];
+
+_actions pushBack [
+    [
+        format["TFAR_External_Intercom_Wireless_Channels", _intercomChannel],
+        localize ELSTRING(Core,Intercom_ACESelfAction_Name),
+        "",
+        {},
+        {
+            params ["_vehicle", "_player"];
```
dedmen, line 57:
```suggestion
            params ["", "_player"];
```

addons/external_intercom/functions/fnc_connect.sqf
```diff
@@ -0,0 +1,208 @@
...
+    [_vehicle, _player] call TFAR_external_intercom_fnc_connect;
+
+  Public: Yes
+*/
+params ["_vehicle", "_player", "_actionParams"];
+_actionParams params [["_wireless", false, [false]]];
+
+if (!alive _vehicle) exitWith {
+    "no";
+};
```
dedmen, line 28:
```suggestion
if (!alive _vehicle) exitWith {"no"};
```

addons/external_intercom/functions/fnc_connect.sqf
```diff
@@ -0,0 +1,208 @@
...
+    _externalSpeakers set [0, _player];
+};
+_vehicle setVariable ["TFAR_ExternalIntercomSpeakers", _externalSpeakers, true];
+
+// Get intercom channel, now that we've signed on to the intercom
+private _intercomSlot = [_vehicle, _player] call TFAR_external_intercom_fnc_getIntercomChannel;
+
+_vehicle setVariable [format ["TFAR_IntercomSlot_%1",(netId _player)], _intercomSlot, true];
+// Because TFAR checks the "vehicle", and thinks the unit is the vehicle when it's not in a vehicle, so we manually set it
+_player setVariable [format ["TFAR_IntercomSlot_%1",(netId _player)], _intercomSlot, true];
```
dedmen, line 80:
```suggestion
_vehicle setVariable [format ["TFAR_IntercomSlot_%1", (netId _player)], _intercomSlot, true];
// Because TFAR checks the "vehicle", and thinks the unit is the vehicle when it's not in a vehicle, so we manually set it
_player setVariable [format ["TFAR_IntercomSlot_%1", (netId _player)], _intercomSlot, true];
```

addons/external_intercom/CustomRope.hpp
```diff
@@ -0,0 +1,23 @@
+class CfgVehicles {
+    class Rope;
+    class RopeSmallWire : Rope {
```
dedmen, line 3: TFAR specific prefix to prevent mod conflicts

addons/external_intercom/functions/fnc_connect.sqf
```diff
@@ -0,0 +1,208 @@
...
+            _vehicle addEventHandler ["RopeBreak", {
+                params ["_vehicle", "_ropeID"];
+                systemChat "Rope broke";
+                systemChat str _this;
+                if ((_vehicle getVariable ["TFAR_ExternalIntercomRopeIDs", [nil, nil]] select 0) isEqualTo _ropeID) then {
+                    [_vehicle, _vehicle getVariable ["TFAR_ExternalIntercomSpeakers", [objNull, []]] select 0] call TFAR_external_intercom_fnc_disconnect;
+                    _vehicle removeEventHandler ["RopeBreak", _thisEventHandler];
+                };
+            }],
+            true
```
dedmen, line 156: whaatt... is this true for? setVariable public with a EH ID? EH ID's are local only?

addons/external_intercom/functions/fnc_connect.sqf
```diff
@@ -0,0 +1,208 @@
...
+        ];
+        diag_log "Registered RopeBreak EH";
+        diag_log (_vehicle getVariable "TFAR_ExternalIntercomRopeEH");
```
dedmen, line 159: missing cleanup.
You can use TRACE_ macros, and enable debug mode based on `__A3_DEBUG__` macro

addons/external_intercom/functions/fnc_disconnect.sqf
dedmen: boop still open, though unlikely to happen. But maybe add a comment that thats possible and could potentially be fixed in the future

addons/external_intercom/functions/fnc_getInteractionPoint.sqf
```diff
@@ -0,0 +1,42 @@
...
+  Public: Yes
+*/
+params ["_vehicle"];
+
+private _customPosition = getArray((configOf _vehicle) >> "TFAR_ExternalIntercomInteractionPoint");
+
+// If there's a custom position, then don't bother calculating
+if !(_customPosition isEqualTo []) exitWith {
+   _customPosition;
+};
```
dedmen, line 28:
```suggestion
if !(_customPosition isEqualTo []) exitWith {_customPosition};
```

addons/external_intercom/functions/fnc_getIntercomChannel.sqf
```diff
@@ -0,0 +1,31 @@
...
+    [_vehicle, _player] call TFAR_fnc_getIntercomChannel;
+
+  Public: Yes
+*/
+params ["_vehicle", "_player"];
+
+// Get any occupants, from crew down to passengers
+private _crew = crew _vehicle;
+_vehicle getVariable [format ["TFAR_IntercomSlot_%1", netId _player],
+    _vehicle getVariable [format ["TFAR_IntercomSlot_%1",(netId (_crew select 0))],
```
dedmen, line 28:
```suggestion
private _crew = crew _vehicle;
_vehicle getVariable [format ["TFAR_IntercomSlot_%1", netId _player],
    _vehicle getVariable [format ["TFAR_IntercomSlot_%1", (netId (_crew select 0))],
```

addons/external_intercom/functions/fnc_parseClassnamesInput.sqf
```diff
@@ -0,0 +1,40 @@
...
+    ["[B_CrewHelmet, 51122, 52, 53, 4, 5, 56, 57, 58, 59, ""asd"", asd, ""88""]", 8, 87, 40] call TFAR_external_intercom_fnc_parseClassnamesInput;
+
+  Public: Yes
+*/
+params [
+    ["_valueString", "", [""]]
+];
+
+private _parsedValue = (_valueString splitString ",") apply {
+    if !(_x isEqualType "STRING") then {_x = format["%1", _x]};
```
dedmen, line 26:
```suggestion
    if !(_x isEqualType "") then {_x = format["%1", _x]};
```
maybe a bit less readable, but.. more efficient? extremely tiny bit? Doesn't really matter

addons/external_intercom/functions/fnc_parseClassnamesInput.sqf
```diff
@@ -0,0 +1,40 @@
...
+    [_x] call BIS_fnc_filterString;
```
dedmen, line 27:
```suggestion
    [_x] call BIS_fnc_filterString
```
if you intend to return this value leave away semicolon on return values

addons/external_intercom/functions/fnc_parseClassnamesInput.sqf
```diff
@@ -0,0 +1,40 @@
...
+};
+
+_parsedValue = _parsedValue arrayIntersect _parsedValue; // Retain only unique elements
+
+_parsedValue = _parsedValue apply {
```
dedmen, line 32:
```suggestion
_parsedValue = _parsedValue apply { // convert to config case
```

addons/external_intercom/functions/fnc_parseClassnamesInput.sqf
```diff
@@ -0,0 +1,40 @@
...
+    if (isClass(configFile >> "CfgWeapons" >> _x)) then {
+        configName(configFile >> "CfgWeapons" >> _x);
+    } else {""};
+};
+
+_parsedValue = _parsedValue select {!(_x isEqualTo "")}; // Retain only non-null elements
```
dedmen, line 38:
```suggestion
_parsedValue = _parsedValue - [""]; // Retain only non-null elements
```

addons/external_intercom/functions/fnc_settingForceAlphaNumericArray.sqf
```diff
@@ -0,0 +1,29 @@
...
+    ["123 123 , 332 23"] call TFAR_external_intercom_fnc_settingForceAlphaNumericArray;
+
+  Public: No
+*/
+params [["_input", "", [""]]];
+
+if (_input isEqualTo "") exitWith {""};
+// keep the last sign, if you want to type a new value
+private _lastSign = _input select [count _input - 1];
+_lastSign = if !((_lastSign isEqualTo ",")||{_lastSign isEqualTo " "}) then {""};
```
dedmen, line 25:
```suggestion
_lastSign = if !((_lastSign isEqualTo ",") || {_lastSign isEqualTo " "}) then {""};
```

addons/external_intercom/script_component.hpp
```diff
@@ -0,0 +1,18 @@
+#define COMPONENT external_intercom
+#include "\z\tfar\addons\core\script_mod.hpp"
+
+#ifdef DEBUG_ENABLED_EXTERNAL_INTERCOM
```
dedmen, line 4: In Arma 2.06 you can set preprocessor macros via start parameter
https://community.bistudio.com/wiki/Arma_3:_Startup_Parameters
preprocDefine
could use this here to dynamically enable intercom debugging

addons/external_intercom/ui/tfar_ace_interaction_external_intercom_phone.svg
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dedmen, line 1: These don't need to be inside a pbo, waste a bit of space
We have other source files in https://github.com/michail-nikolaev/task-force-arma-3-radio/tree/master/extra
but this is small enough to not matter

addons/external_intercom/ACEActions.hpp
```diff
@@ -0,0 +1,78 @@
+#define ADD_PHONE_ACTIONS \
+class TFAR_External_Intercom_Phone_Base { \
```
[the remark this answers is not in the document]
Saborknight: I see what you're saying and that does make sense, but from the player's perspective, having to go through an extra menu item when interacting with a vehicle that could potentially be moving at the time would probably be difficult to do. Even if not moving, it's still an extra step that UX wise doesn't make sense....
Is there a way perhaps to still make it more efficient without making children? Or do some quick replacement maybe?.... maybe I could add the interactions via script instead, and have an overall check there to see if the vehicle even has intercom. What do you think @dedmen ?

addons/external_intercom/CustomInteractionPoints.hpp
Saborknight: Good point!

addons/external_intercom/XEH_preInit.sqf
Saborknight: Well.... in testing it has. Why wouldn't it? @dedmen

addons/external_intercom/functions/fnc_parseClassnamesInput.sqf
Saborknight: And does this apply to the end of file lines too? End of functions, where it outputs the function return?

addons/external_intercom/script_component.hpp
Saborknight: Not too sure how that all works? This was just a copy/paste from one of the other addons

Saborknight: Oh, I see what you mean. You're saying I can set this variable to true in my launcher for additional logging. Thanks for the heads up!

addons/intercom_phone/functions/fnc_getIntercomChannel.sqf
Saborknight: Seems to work fine

addons/external_intercom/CfgVehicles.hpp
```diff
@@ -1,41 +1,17 @@
...
 
-#define LAND_CLASS(parent, child) \
-class child : parent { \
-    class ACE_Actions { \
-        ADD_PHONE_ACTIONS \
-        ADD_WIRELESS_ACTIONS \
-    }; \
-};
+#define LAND_INTERACTIONS \
+init = QUOTE(call FUNC(addWirelessInteractions);call FUNC(addPhoneInteractions));
```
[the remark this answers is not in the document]
Saborknight: Ah well, in that case I'll move to that instead. Thanks!

nomisum:
eagerly waiting for this, want to use that rope and model as a dependency for a telephone mod :) 

love the external intercom feature too.

addons/external_intercom/CfgVehicles.hpp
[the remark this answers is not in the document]
Saborknight: > 

So I've tested it, and it doesn't seem to execute on an object unless I actually mount into it. So, this (using the `newControllableObject` eventhandler) isn't going to work out.

addons/external_intercom/functions/fnc_connect.sqf
nomisum: Did you figure out how/when _ropeBreak_ works? Seems to work for direct connect to units for me but as soon as the ropeAttached object is attached itself, it wont. Still investigating if a workaround is possible.

Saborknight: Nope, though if what you say is the case... maybe a base-game bug or something? Or maybe the eventhandler needs to be applied to something/one else? I would have thought the vehicle were enough, but maybe it needs to be applied to the phone too?

nomisum: I am working on a phone mod and attach the rope from helperobject (heliclass) to player (lefthand selection) and with this setup ropeBreak works. you need to add the EH to the helperobject (heli class).

Saborknight:
Is this waiting on anything else? I can't really make any more changes past the 26th as I'm hoping to be in the military and therefore won't be available at that point. So, anything that you guys need from me, please let me know beforehand. Thanks

dedmen:
Ugh...
Many build errors.
ACEActions.hpp has two TFAR_Radios classes, that can't be.
CustomInteractionPoints.hpp has multiple base class declarations for B_APC_Tracked_01_base_F, that also can't be

dedmen:
```
18:51:00 Unexpected stringtable format inside <Text ID="STR_TFAR_EXTERNAL_INTERCOM_SOMEONE_USING_PHONE"><Original>
18:51:00 Unexpected stringtable format inside <Text ID="STR_TFAR_EXTERNAL_INTERCOM_SOMEONE_USING_PHONE"><English>
18:51:00 Unexpected stringtable format inside <Text ID="STR_TFAR_EXTERNAL_INTERCOM_SOMEONE_USING_PHONE"><Czech>
18:51:00 Unexpected stringtable format inside <Text ID="STR_TFAR_EXTERNAL_INTERCOM_SOMEONE_USING_PHONE"><Russian>
18:51:00 Unexpected stringtable format inside <Text ID="STR_TFAR_EXTERNAL_INTERCOM_SOMEONE_USING_PHONE"><Italian>
18:51:00 Unexpected stringtable format inside <Text ID="STR_TFAR_EXTERNAL_INTERCOM_HEADGEAR_NOT_WIRELESS_CAPABLE"><Original>
18:51:00 Unexpected stringtable format inside <Text ID="STR_TFAR_EXTERNAL_INTERCOM_HEADGEAR_NOT_WIRELESS_CAPABLE"><English>
18:51:00 Unexpected stringtable format inside <Text ID="STR_TFAR_EXTERNAL_INTERCOM_HEADGEAR_NOT_WIRELESS_CAPABLE"><Czech>
18:51:00 Unexpected stringtable format inside <Text ID="STR_TFAR_EXTERNAL_INTERCOM_HEADGEAR_NOT_WIRELESS_CAPABLE"><Russian>
18:51:00 Unexpected stringtable format inside <Text ID="STR_TFAR_EXTERNAL_INTERCOM_HEADGEAR_NOT_WIRELESS_CAPABLE"><Italian>
```


😢

PR #1531, opened 2021-07-21
Fix building with latest mikero tools

Dahlgren:
- Fix building with latest mikero tools


addons/core/intercomConfig/config.cpp
```diff
@@ -8,7 +8,7 @@ class CfgPatches {
         units[] = {};
         weapons[] = {};
         requiredVersion = REQUIRED_VERSION;
-        requiredAddons[] = {};
+        requiredAddons[] = {"A3_Data_F"};
```
dedmen, line 11: tfar_core

tools/make.py
```diff
@@ -1243,10 +1235,7 @@ def main(argv):
                         cmd = [makepboTool, "-P","-A","-L","-G","-X=*.backup", os.path.join(work_drive, prefix, module),os.path.join(module_root, release_dir, project,"addons")]
 
                     else:
-                        if check_external:
-                            cmd = [pboproject, "-P", os.path.join(work_drive, prefix, module), "+Engine=Arma3", "-S","+Noisy", "+X", "+Clean", "+Mod="+os.path.join(module_root, release_dir, project), "-Key"]
-                        else:
-                            cmd = [pboproject, "-P", os.path.join(work_drive, prefix, module), "+Engine=Arma3", "-S","+Noisy", "-X", "+Clean", "+Mod="+os.path.join(module_root, release_dir, project), "-Key"]
+                        cmd = [pboproject, "-P", os.path.join(work_drive, prefix, module), "+Engine=Arma3", "-S","+Noisy", "+Clean", "+Mod="+os.path.join(module_root, release_dir, project), "-Key"]
```
dedmen, line 1238: So this is checkexternal disabled? as this runs on CI with no pdrive

addons/core/intercomConfig/config.cpp
dedmen:
```suggestion
        requiredAddons[] = {"tfar_core"};
```

tools/make.py
Dahlgren: Yes, `+X` and `-X` no longer exists 😞

addons/core/intercomConfig/config.cpp
Dahlgren: Fixed

tools/make.py
dedmen: Well then this won't work.
CI uses older mikeros tools anyway

PR #1545, opened 2022-04-17
Workaround for guer to independent in getSideRadio

### LINKED ISSUE #1539: Side Default Radio for Independent uses the wrong str() value
`str(independent)` in TFAR_fnc_getSideRadio returns "GUER", however the init variables from TFAR create "TFAR_DefaultRadio_Personal_Independent", causing retrieval of these variables based on string interpolation to fail. This requires compensation code to work around.

```sqf
missionNamespace setVariable [
  "TFAR_DefaultRadio_Personal_Guer",
  TFAR_DefaultRadio_Personal_Independent,
  true
];
missionNamespace setVariable [
  "TFAR_DefaultRadio_Backpack_Guer",
  TFAR_DefaultRadio_Backpack_Independent,
  true
];
```

https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/c408ce13e9300069207213fb75df49f0b5dfe7c5/addons/core/functions/fnc_getSideRadio.sqf#L26-L37

Dahlgren:
- Fix #1539

### DIFF
--- a/addons/core/functions/fnc_getSideRadio.sqf
+++ b/addons/core/functions/fnc_getSideRadio.sqf
@@ -23,6 +23,10 @@
 
 params ["_side", "_radioType"];
 
+if (_side == independent) then {
+    _side = "Independent";
+};
+
 private _result = "";
 switch (_radioType) do {
     case 0: {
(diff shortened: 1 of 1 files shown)


addons/core/functions/fnc_getSideRadio.sqf
```diff
@@ -23,6 +23,10 @@
 
 params ["_side", "_radioType"];
 
+if (_side == independent) then {
```
dedmen, line 26:
```suggestion
if (_side == independent) then { // hotfix, we want "independent" instead of "guer"
```

PR #1551, opened 2022-10-25
Fix hearing alive players from spectator when its disabled

smitt14ua:
Fix hearing alive players (enemies/friendlies) when its disabled via settings.
Reason was wrong logic formula for isNotHearableInNonPureSpectator variable.

### DIFF
--- a/addons/core/functions/plugin/fnc_sendPluginConfig.sqf
+++ b/addons/core/functions/plugin/fnc_sendPluginConfig.sqf
@@ -19,7 +19,11 @@
 */
 
 
-["minimumPluginVersion", 327] call TFAR_fnc_setPluginSetting; //Keep this as first
+if !(missionNamespace getVariable ["TFAR_spectatorCanHearFriendlies",true]) then {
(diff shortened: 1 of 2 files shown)


Canibus:
would be great if @dedmen will update with this request a parameter "minimumPluginVersion" to 331 also.

dedmen (approved):

dedmen:
Ah I'm not sure about that.
Forcing everyone to update plugin, even if not using this..

How about
https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/master/addons/core/functions/plugin/fnc_sendPluginConfig.sqf#L22

```sqf
if !(missionNamespace getVariable ["TFAR_spectatorCanHearFriendlies",true]]) then
{
    ["minimumPluginVersion", 331] call TFAR_fnc_setPluginSetting; //Temporary until next plugin update. This is only needed when using non-default settings for spectators
} else {
    ["minimumPluginVersion", 327] call TFAR_fnc_setPluginSetting; //Keep this as first
};
```
?
I think thats good, can you add that to the PR?

Canibus:
Could you publish 331 build ?

Issue #1560, opened 2023-06-27
Spectators moved into serious channel - noAutoMoveSpectator is never sent to plugin


ilbinek:
**Arma 3 Version:** `2.12` (stable)
**CBA Version:** `3.15.8` (stable)
**Task Force Radio Version:** `stable` (stable)

**Mods:**
```
- @CBA_A3
- @task_force_radio
```
**Description:**
- The player gets moved into the serious channel even when he is in the spectator interface and noAutomoveSpectaor is true

**Steps to reproduce:**
- Set "noAutomoveSpectaor" to true
- Kill yourself
- Change channel

**Where did the issue occur?**
- Dedicated / Self-Hosted Multiplayer / Editor (Multiplayer)

**Cause**
This is caused, because "noAutoMoveSpectator" is missing in https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/master/addons/core/functions/plugin/fnc_sendPluginConfig.sqf

So the only time the setting is sent to the TS plugin is during init and that is stopped here https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/02b1b0cd6b3eda8ebc10137134b78ca71862b9c6/addons/core/functions/plugin/fnc_setPluginSetting.sqf#L23

This results in the TS plugin never receiving the "true" and staying "false" by default.

PR #1561, opened 2023-06-27
Add noAutoMoveSpectator into config that is sent to plugin

ilbinek:
noAutoMoveSpectator is not sent to plugin. Players will no longer be moved into serious channel from spectator.

- sends the value of noAutomoveSpectator to the plugin

Solves #1560


dedmen:
Will you edit this to add the debug thing?

PR #1571, opened 2024-11-27
Intercom - Add functions and API event

DartRuffian:
- Add dedicated functions for the condition and statement of intercom channels
- Add event for when an intercom channel is set

### DIFF
--- a/addons/core/functions/intercom/fnc_getIntercomChannel.sqf
+++ b/addons/core/functions/intercom/fnc_getIntercomChannel.sqf
@@ -1,0 +1,33 @@
+#include "script_component.hpp"
+
+/*
+  Name: TFAR_fnc_getIntercomChannel
+
+  Author: DartRuffian
+    Gets the channel of a player's vehicle intercom.
+
+  Arguments:
+    0: Vehicle <OBJECT>
+    1: Player <OBJECT>
+
+  Return Value:
+    The intercom channel <NUMBER>
+
+  Example:
+    [vehicle player, player] = call TFAR_fnc_getIntercomChannel;
+
+  Public: Yes
+*/
+
+params [
+    ["_vehicle", objNull, [objNull]],
+    ["_player", objNull, [objNull]]
+];
+
+if (!isPlayer _player) exitWith {};
+
+private _intercom = _vehicle getVariable [format ["TFAR_IntercomSlot_%1", netId _player], -2];
+if (_intercom == -2) then {
+    _intercom = _vehicle getVariable ["TFAR_defaultIntercomSlot", TFAR_defaultIntercomSlot];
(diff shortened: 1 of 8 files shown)


DartRuffian:
@dedmen

addons/core/functions/intercom/fnc_setIntercomChannel.sqf
```diff
@@ -0,0 +1,33 @@
...
+params [
+    ["_vehicle", objNull, [objNull]],
+    ["_player", objNull, [objNull]],
+    ["_channel", 1, [1]]
+];
+
+if (!alive _vehicle || !alive _player || !isPlayer _player) exitWith {};
+
+_vehicle setVariable [format ["TFAR_IntercomSlot_%1", netId _player], _channel, true];
+["TFAR_intercomChannelSet", [_vehicle, _player, _channel]] call CBA_fnc_localEvent;
```
dedmen, line 32: Usually we use tfars event system. Which is a wrapper around CBA.
I don't know if we do it like this elsewhere, it's probably fine?

dedmen (approved):

addons/core/functions/intercom/fnc_setIntercomChannel.sqf
DartRuffian: I saw fnc_fireEventHandlers but the function description was "Fires all eventhandlers associated with the passed unit", which seemed to imply it was mainly for units, so I left it as just a CBA event directly.

DartRuffian: I changed it because I realized that all other TFAR events are `TFAR_event_<name>`, so I just changed it to use the fireEventHandlers function for consistency.

All other events are also named `onX`, so I changed the event name to `onIntercomChannelSet` to match.

addons/core/functions/intercom/fnc_getIntercomChannel.sqf
```diff
@@ -0,0 +1,33 @@
...
+
+  Public: Yes
+*/
+
+params [
+    ["_vehicle", objNull, [objNull]],
+    ["_player", objNull, [objNull]]
+];
+
+if (isNull _vehicle || !isPlayer _player) exitWith {};
```
DartRuffian, line 27: Should this return `-1` or `-2` on early exit?

Returning a non-nil value would make it so third party mods don't need to explicitly check the player's vehicle, and could just call the function.
```sqf
if ([objectParent _unit, _unit] call TFAR_fnc_getIntercomChannel) >= 0) then {
    // _unit is a player, and is on an intercom
} else {
    // _unit is either not a player, or is not in a vehicle
};
```

PR #1586, opened 2026-08-28
Add hemtt support

ilbinek:
- change the build toolchain from mikero tools to hemtt

**Limitation**
- hemtt doesn't allow the usage of "-" in version number - currently set to "1.1.0.342"


.hemtt/project.toml
```diff
@@ -0,0 +1,36 @@
...
+[hemtt.release]
+sign = true
+archive = false
+
+[hemtt.launch.default]
+workshop = [
+    "450814997", # CBA_A3
+]
+
+[hemtt.launch.ace]
```
DartRuffian, line 31: Can just use extends = "default" and remove cba

addons/external_intercom/functions/fnc_addWirelessIntercomMenu.sqf
```diff
@@ -30,7 +30,7 @@ private _actions = [];
 
 _actions pushBack [
     [
-        format ["TFAR_External_Intercom_Wireless_Disconnect", _intercomChannel],
+
```
DartRuffian, line 33: Empty line

DartRuffian (commented):
Should add a hemtt check workflow for prs

addons/external_intercom/functions/fnc_addWirelessIntercomMenu.sqf
SpicyBagpipes: tried building without that empty line and it crashed everything.

DartRuffian: The m in KJW stands for misinformation

.hemtt/project.toml
```diff
@@ -0,0 +1,36 @@
...
+workshop = [
+    "450814997", # CBA_A3
+    "463939057", # ACE
+    "3499977893", # Advanced Developer Tools
```
dedmen, line 35: Why? ADT is not required

.hemtt/project.toml
```diff
@@ -0,0 +1,36 @@
+name = "Task Force Arrowhead Radio"
+author = "Nkey, Deadmen"
```
dedmen, line 2: What is this used for?
We have https://github.com/michail-nikolaev/task-force-arma-3-radio/blob/master/addons/core/stringtable.xml#L5

addons/core/script_version.hpp
```diff
@@ -1,4 +1,4 @@
 #define MAJOR 1
-#define MINOR -1
```
dedmen, line 2: No, if HEMTT cannot handle our versioning, then we turn off HEMTT's versioning, and let the CI scripts deal with writing this file. (The CI scripts already write this file, but I fear HEMTT will try to overwrite it, after we already put in what we need...)

addons/external_intercom/functions/fnc_connect.sqf
```diff
@@ -32,15 +32,15 @@ if (TFAR_externalIntercomEnable isEqualTo 2) exitWith {"no"};
 if (TFAR_externalIntercomEnable isEqualTo 1 && [side _player, side _vehicle] call BIS_fnc_sideIsEnemy) exitWith {"no"};
 
 if (_wireless && !([_player] call FUNC(hasWirelessHeadgear))) exitWith {
-    [parseText format [CSTRING(HEADGEAR_NOT_WIRELESS_CAPABLE), getText(configFile >> "CfgWeapons" >> headgear _player >> "displayName")]] call TFAR_fnc_showHint;
+    [parseText format [localize LSTRING(HEADGEAR_NOT_WIRELESS_CAPABLE), getText(configFile >> "CfgWeapons" >> headgear _player >> "displayName")]] call TFAR_fnc_showHint;
```
dedmen, line 35: localize LSTRING, has a macro `LLSTRING` https://github.com/CBATeam/CBA_A3/blob/master/addons/main/script_macros_common.hpp#L1279

Curious how this code worked before, without calling localize... Guess it never did?

dedmen:
I don't like renaming SQF files containing SQF scripts that get included into other SQF files, to .hpp
Didn't hemtt have support .inc.sqf extension? Keeping syntax highlighting while avoiding false macro errors?

Besides that, and the open comments, everything else is logical 👍

addons/core/script_version.hpp
dedmen: If this is only about HEMTT not allowing CfgPatches "version" entry with it.
Then, just keep it as is, latest release has
`version="1.-1.0.341"`

addons/core/script_mod.hpp
```diff
@@ -5,11 +5,13 @@
 
 #include "script_version.hpp"
 
-#define VERSION MAJOR.MINOR.PATCHLVL.BUILD
```
dedmen, line 8: This changes CfgPatches version entry from
`version="1.-1.0.341";`
to
`version=1.-1`

That is both an unnecessary change, and invalid syntax, and HEMTT would block it
Why change it from string to number? Just keep it a string.

addons/external_intercom/functions/fnc_addWirelessIntercomMenu.sqf
PabstMirror:
```suggestion
          "TFAR_External_Intercom_Wireless_Disconnect"
```
hemtt doesn't like the format without a %1 for the arg

addons/external_intercom/XEH_preStart.sqf
```diff
@@ -1,3 +1,3 @@
 #include "script_component.hpp"
 
-#include "XEH_PREP.sqf"
+#include "XEH_PREP.hpp"
```
PabstMirror, line 3: `.inc.sqf` might be more commonly used

.hemtt/project.toml
```diff
@@ -0,0 +1,36 @@
...
+workshop = [
+    "450814997", # CBA_A3
+]
+
+[hemtt.launch.ace]
+workshop = [
+    "450814997", # CBA_A3
+    "463939057", # ACE
+    "3499977893", # Advanced Developer Tools
+]
```
PabstMirror, line 36:
```suggestion
]


[lints.sqf.var_all_caps]
options.ignore = [
    "TF_MAX_LR_VOLUME", "TF_MAX_SW_VOLUME", "TF_UNDERWATER_RADIO_DEPTH"
]

# These are good lints, but may want to disable at first while dealing with more important issues
[lints.config.file_type]
enabled = "pedantic"
[lints.sqf.command_case]
enabled = "pedantic"
[lints.sqf.extra_not]
enabled = "pedantic"
[lints.sqf.config_of]
enabled = "pedantic"
[lints.sqf.if_assign]
enabled = "pedantic"
[lints.sqf.in_vehicle_check]
enabled = "pedantic"

[inspector]
header_regex = "ace"
vars_to_ignore = []
check_function_calls = true

```

.hemtt/project.toml
ilbinek: it is not, it's a personal preference

doing anything where console is probably gonna be used, it's a must have for me

guess I could add an `ace_adt` config

addons/external_intercom/functions/fnc_addWirelessIntercomMenu.sqf
ilbinek: fixed, my mistake

ilbinek:
includes https://github.com/michail-nikolaev/task-force-arma-3-radio/pull/1587 already, so could be closed

.hemtt/project.toml
ilbinek: removed ADT from ace preset, but added ace_adt for personal convinience.

.hemtt/project.toml
ilbinek: I'll be honest, no idea. But reaplced it with "Task Force Arrowhead Radio Team"

addons/core/script_version.hpp
ilbinek: pain, but got it working

addons/external_intercom/functions/fnc_connect.sqf
ilbinek: changed to LLSTRING

looking through the code, there's another 113 "localize LSTRING" entries. Something to _fix_ sometime, but won't clutter this PR with it.
image

addons/core/script_mod.hpp
ilbinek: fixed to be compatible

addons/external_intercom/XEH_preStart.sqf
ilbinek: ^^ renamed

ilbinek:
@dedmen I reverted the deletion of the old build scripts. I am unsure how exactly your build process works, so I got them back in.

Only drawback I can see now, is that if you `hemtt release` the bikey bisign files wll be `*_1.0.0.bisign`. But I assume you handle those yourself during builds.

dedmen:
The build process handles the signing separately yeah. So shouldn't matter what hemtt does, besides me having to delete what hemtt did before I redo my own ontop of that.
I don't even remember what scripts the build process uses or not, but I have to rebuild it for HEMTT anyway.

.hemtt/project.toml
dedmen: Those are in lints.toml  now, so this is resolved right?

addons/core/functions/events/handler/fnc_addEventHandler.sqf
```diff
@@ -13,7 +13,7 @@
     3: Unit to use as filter. <OBJECT> (default: nil)
 
   Return Value:
-    Unique ID of the event handler <NUMBER>
+    Events (CBA Hash) <ARRAY>
```
dedmen, line 16: This was originally supposed to be
`NUMBER: Unique ID of the event handler (can be used with CBA_fnc_removeEventHandler).`
to return _eventID, which can then be passed also into TFAR_fnc_removeEventHandler.

There is not much sense in returning the CBA hash

This should stay number, and instead add `_eventID` at the end of file.

dedmen (approved):

.hemtt/project.toml
ilbinek: from my understanding of the hemtt book, yes, better to not clutter project.toml with it in my opinion

addons/core/functions/events/handler/fnc_addEventHandler.sqf
dedmen:
```suggestion
    Unique ID of the event handler (can be used with CBA_fnc_removeEventHandler) <NUMBER>
```

dedmen:
Just that one thing with fnc_addEventHandler returning the _eventId 🥺

addons/core/functions/events/handler/fnc_addEventHandler.sqf
PabstMirror, line 57:
```suggestion
//We were editing _handlersHash by reference so we don't call setVariable again

_eventID

```

